PR #4, opened 2019-09-27
[WIP] feat: use requesting pod's service account token, not driver service account token

malnick:


mikesir87:
I would _love_ to see this feature included, as this is one of our biggest barriers to using the CSI driver. It only takes one configuration oops to give secrets to an app that shouldn't have them. Any update on progress or is there anywhere else discussion might be happening?

tomhjp:
Added in #64

mikesir87:
Excellent! Thank you @tomhjp! 🎉

PR #6, opened 2019-10-16
Generates secret endpoint

carnei-ro:
Created a function that generates the endpoint for the KV/2 secret.
Removed hard coded secret url.


ritazh:
fixes https://github.com/deislabs/secrets-store-csi-driver/issues/91

carnei-ro:
Overlaps PR #7

PR #8, opened 2019-10-17
Support KV/1, KV/2

carnei-ro:
Dynamic creates secret endpoint based on "objectPath".
Needs "read" permission on "/sys/mounts" to discovery secret mount
 type and version.
Writes to objectName instead of concatenate path to support nested
 secret path.

Fixes deislabs/secrets-store-csi-driver#91


carnei-ro:
@anubhavmishra @ritazh

malnick:
@carnei-ro Thanks for the contribution! Sorry for the delay in review, we will get this manually tested (in lieu of having an e2e test in CI right now) and reviewed this week.

provider.go
```diff
@@ -77,6 +83,63 @@ func readJWTToken(path string) (string, error) {
 	return string(bytes.TrimSpace(data)), nil
 }
 
+func splitSecretPath(secretPath string) (string, string) {
+	s := regexp.MustCompile("/+").Split(secretPath, 3)
+	return s[1], s[2]
```
kalafut: `len(s)` could possibly be less than 3. Given the small size of this function and its single use, it may be appropriate to move the `regexp` call to the calling site and check the result length there (since you'll be able to return an error).

provider.go
```diff
@@ -77,6 +83,63 @@ func readJWTToken(path string) (string, error) {
...
+}
+
+func (p *Provider) getMountInfo(mountName string, token string) (string, string, error) {
```
kalafut:
```suggestion
func (p *Provider) getMountInfo(mountName, token string) (string, string, error) {
```

provider.go
```diff
@@ -77,6 +83,63 @@ func readJWTToken(path string) (string, error) {
...
+	client, err := p.createHTTPClient()
+	if err != nil {
+		return "", "", err
+	}
+
+	addr := p.VaultAddress + "/v1/sys/mounts"
+	req, err := http.NewRequest(http.MethodGet, addr, nil)
+	// Set vault token.
+	req.Header.Set("X-Vault-Token", token)
```
kalafut: Also set `X-Vault-Request` to `"true"`. This is a [recently-added SSRF mitigation](https://github.com/hashicorp/vault/blob/14c7398841a1b0794a42882f7a6d223d6cbecf59/vendor/github.com/hashicorp/vault/api/client.go#L438).

provider.go
```diff
@@ -77,6 +83,63 @@ func readJWTToken(path string) (string, error) {
...
+	if err != nil {
```
kalafut: This check should be immediately following `http.NewRequest`.

provider.go
```diff
@@ -77,6 +83,63 @@ func readJWTToken(path string) (string, error) {
...
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&mount); err != nil {
+		fmt.Printf("error: %s", err)
+	}
+
+	return mount.Data[mountName+"/"].Type, mount.Data[mountName+"/"].Options["version"], nil
+}
+
+func generateSecretEndpoint(vaultAddress string, secretMountType string, secretMountVersion string, secretPrefix string, secretSuffix string, secretVersion string) string {
```
kalafut: I'd suggest returning an error here, in particular explicit errors if `secretMountType != "kv"` or the version isn't 1 or 2.

carnei-ro:
Thanks for the review guys, changes has been made!

provider.go
```diff
@@ -77,6 +83,64 @@ func readJWTToken(path string) (string, error) {
...
+		return "", "", fmt.Errorf("failed to get successful response: %#v, %s",
+			resp, b.String())
+	}
+
+	var mount struct {
+		Data map[string]Mount `json:"data"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&mount); err != nil {
+		fmt.Printf("error: %s", err)
```
ritazh: Should return the err here after logging

provider.go
```diff
@@ -228,14 +229,22 @@ func (p *Provider) getSecret(token string, secretPath string, secretName string,
 		secretVersion = "0"
 	}
 
-	secretPrefix, secretSuffix := splitSecretPath(secretPath)
+	secretPrefix, secretSuffix := "", ""
```
kalafut: Instead of defaults, I think it would be better to fail if we don't have a prefix/suffix. Something like:

```go
s := regexp.MustCompile("/+").Split(secretPath, 3)
if len(s) < 3 {
	return "", fmt.Errorf("unable to parse secret path %q", secretPath)
}
secretPrefix := s[1]
secretSuffix := s[2]
```

provider.go
```diff
@@ -342,10 +436,10 @@ func (p *Provider) MountSecretsStoreObjectContent(ctx context.Context, attrib ma
 			return err
 		}
 		objectContent := []byte(content)
-		if err := ioutil.WriteFile(path.Join(targetPath, keyValueObject.ObjectPath), objectContent, permission); err != nil {
-			return errors.Wrapf(err, "secrets-store csi driver failed to write %s at %s", keyValueObject.ObjectPath, targetPath)
+		if err := ioutil.WriteFile(path.Join(targetPath, keyValueObject.ObjectName), objectContent, permission); err != nil {
+			return errors.Wrapf(err, "secrets-store csi driver failed to write %s at %s", keyValueObject.ObjectName, targetPath)
```
ritazh: Changing from `ObjectPath` to `ObjectName` changes the existing behavior of this provider. Currently when user deploys this:

```yaml
apiVersion: secrets-store.csi.k8s.com/v1alpha1
kind: SecretProviderClass
metadata:
  name: vault-foo
spec:
  provider: vault
  parameters:
    roleName: "example-role"
    vaultAddress: http://${VAULT_SERVICE_IP}:8200
    vaultSkipTLSVerify: "true"
    objects:  |
      array:
        - |
          objectPath: "/foo"
          objectName: "bar"
          objectVersion: ""
```
From the container that uses the existing Vault provider, it will expect the value of `bar` to be written into `mountPath: "/mnt/secrets-store"` at `/mnt/secrets-store/foo`.

This change will change this behavior into the value of `bar` is written in `/mount/secrets-store/bar`.

provider.go
```diff
@@ -77,6 +83,65 @@ func readJWTToken(path string) (string, error) {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", "", errors.Wrapf(err, "couldn't get sys mounts")
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		var b bytes.Buffer
+		io.Copy(&b, resp.Body)
```
kalafut: Handle the error returned from `io.Copy`

carnei-ro:
I have coded a little bit more and know it supports KV/1, KV/2 and Database. May I submit in this PR?

provider.go
kalafut: I think this is a reasonable change. It not only avoids the issue of trying to write nested keys to a directory that may not exist, but also lets someone access multiple values within a single key (e.g both "bar" and "baz" within the "/foo" key). Currently the last object in the array would win.

If the path aspect is desirable, we could look at a separate enhancement to prefix the output filename in a manner similar to [envconsul](https://github.com/hashicorp/envconsul#vault). This would result in a filename like "foo_bar" or "foo_subkey_bar". This is could be an opt-in enhancement.

ritazh: SGTM. 
@carnei-ro can you pls add some notes about this change in the readme as this is a breaking change. 

@kalafut @aramase after a new release of the vault provider is cut and updated in the driver, we need to update the e2e test in the driver repo to reflect this change as the e2e test as of now will fail.

provider.go
```diff
@@ -78,6 +84,65 @@ func readJWTToken(path string) (string, error) {
...
+		return "", "", fmt.Errorf("failed to get successful response: %#v, %s",
+			resp, b.String())
+	}
+
+	var mount struct {
+		Data map[string]Mount `json:"data"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&mount); err != nil {
+		fmt.Printf("error: %s", err)
```
aramase: nit: is this debug log? Maybe it should be removed.

kalafut:
Thanks!

PR #19, opened 2019-12-08
switch from glog to logrus

aramase:
This is required to print the `stdout` and `stderr` in the driver logs instead of having a side car container to tail the logs. 

fixes https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/issues/13

We need a new release of `provider-vault` once these changes are merged. Once we have a new release we can cut a new release for driver.


aramase:
cc @ritazh @malnick @kalafut

deployment/provider-vault-installer.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       tolerations:
       containers:
         - name: provider-vault-installer
-          image: hashicorp/secrets-store-csi-driver-provider-vault:0.0.1
+          image: hashicorp/secrets-store-csi-driver-provider-vault:0.0.2
```
ritazh: Should we update this file _after_ a new vault provider `0.0.3` has been released? I think it's currently at `0.0.2`

aramase: It is currently at 0.0.2 and the deployment yaml wasn't up to date. So updated it to reflect the latest. Once `0.0.3` is cut, then this will need to be updated to show the latest image.

ritazh: okay cuz https://github.com/deislabs/secrets-store-csi-driver/blob/master/deploy/provider-vault.yaml#L21-L54 is  at `0.0.2` and it still has the provider-log stuff. I believe `0.0.2` still looks for that volume.

aramase: I was hoping @malnick can get a new release once this PR is merged and we can just bump the version for the image. But I can add remove the `provider-log` when we have a new release.

PR #23, opened 2019-12-12
add --version flag

aramase:
fixes https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/issues/18

```
root@k8s-agentpool1-32372105-vmss000000:/etc/kubernetes/secrets-store-csi-providers/vault# ./provider-vault --version
{"version":"addversion02","buildDate":"2019-12-11-20:35","minDriverVersion":"v0.0.8"}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,9 @@ var (
 	targetPath = pflag.String("targetPath", "", "Target path to write data.")
 	permission = pflag.String("permission", "", "File permission")
 	debug      = pflag.Bool("debug", false, "sets log to debug level")
+	version    = pflag.Bool("version", false, "prints the version information")
+
+	minDriverVersion = "v0.0.8"
 )
 
 // LogHook is used to setup custom hooks
@@ -33,6 +36,13 @@ func main() {
 	var err error
 
 	setupLogger()
+
+	if *version {
+		if err = printVersion(); err != nil {
+			log.Fatalf("failed to print version, err: %+v", err)
+		}
+		os.Exit(0)
+	}
 
 	err = json.Unmarshal([]byte(*attributes), &attrib)
 	if err != nil {
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -18,9 +19,9 @@ var (
 	permission = pflag.String("permission", "", "File permission")
 	debug      = pflag.Bool("debug", false, "sets log to debug level")
 	version    = pflag.Bool("version", false, "prints the version information")
-
-	minDriverVersion = "v0.0.8"
 )
+
+const minDriverVersion = "v0.0.8"
 
 // LogHook is used to setup custom hooks
 type LogHook struct {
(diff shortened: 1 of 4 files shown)


aramase:
cc @malnick @kalafut @ritazh

main.go
```diff
@@ -17,6 +17,9 @@ var (
 	targetPath = pflag.String("targetPath", "", "Target path to write data.")
 	permission = pflag.String("permission", "", "File permission")
 	debug      = pflag.Bool("debug", false, "sets log to debug level")
+	version    = pflag.Bool("version", false, "prints the version information")
+
+	minDriverVersion = "v0.0.8"
```
kalafut: This can be `const`.

main.go
```diff
@@ -34,6 +37,13 @@ func main() {
 
 	setupLogger()
 
+	if *version {
+		if err = printVersion(); err != nil {
```
kalafut:
```suggestion
		if err := printVersion(); err != nil {
```

version.go
```diff
@@ -0,0 +1,37 @@
...
+
+var (
+	// BuildDate is date when binary was built
+	BuildDate string
+	// BuildVersion is the version of binary
+	BuildVersion string
+)
+
+// providerVersion holds current provider version
+type providerVersion struct {
```
kalafut: Can this instead be a type exported from the CSI driver? Seems a little odd for each provider to create it, while still needing to conform to a standard format.

version.go
```diff
@@ -0,0 +1,37 @@
...
+	Version   string `json:"version"`
+	BuildDate string `json:"buildDate"`
+	// MinDriverVersion is minimum driver version the provider works with
+	MinDriverVersion string `json:"minDriverVersion"`
+}
+
+func printVersion() (err error) {
```
kalafut: Generally prefer to avoid the named/naked returns until they're specifically needed, which it doesn't seem like they are here.

version.go
```diff
@@ -0,0 +1,37 @@
...
+		BuildDate:        BuildDate,
+		MinDriverVersion: minDriverVersion,
+	}
+
+	var res []byte
+	if res, err = json.Marshal(pv); err != nil {
+		return
+	}
+
+	fmt.Printf(string(res) + "\n")
```
kalafut:
```suggestion
	fmt.Printf("%s\n", res)
```

version.go
kalafut: Consider passing in an `io.Writer`, which you can the `fmt.Fprintf` to. This is generally a useful decoupling, and will allow us to avoid all of the  special handling of Stdout in the test.

main.go
aramase: done!

main.go
aramase: the error var is declared before this, so this wouldn't be required. But I changed it.

version.go
aramase: done!

version.go
aramase: done!

version.go
aramase: returning `(string,error)`

version.go
aramase: @kalafut I don't think we want to add driver as a dependency here for the version type. WDYT?

kalafut: Yeah I've had a change of heart as this is such a simple structure. Let's leave as is.

main.go
kalafut: I prefer to not shadow anything (usually `err`) unless necessary. Subtle bugs...

aramase: got it!

PR #24, opened 2020-01-10
Update deploy to 0.0.4

ritazh:


deployment/provider-vault-installer.yaml
```diff
@@ -15,7 +18,7 @@ spec:
       tolerations:
       containers:
         - name: provider-vault-installer
-          image: hashicorp/secrets-store-csi-driver-provider-vault:0.0.2
+          image: hashicorp/secrets-store-csi-driver-provider-vault:0.0.4
```
kalafut: Why not 0.0.5?

ritazh: Good question! Per sig-auth recommendation, we are removing providers from the driver deployment to ensure the driver and the providers can work independently. As a result, we are referencing each provider's deployment yaml directly in our e2e test. So this PR is to get all of that setup between the repos.

As you recall 0.0.5 introduced a change that would require us to update our e2e test for vault in the driver, which we should do after this is done so that we can separate the concerns.

HTH

kalafut: Gotch

kalafut:
LGTM!

PR #45, opened 2020-07-02
Add objectAlias support for filenames

### LINKED ISSUE #44: The previous secret will be overwritten by same ObjectName
Tested revision: 34a978c51db6b048873167c9e2003c145d2f5f23

If the same ObjectName is included in a SecretProviderClass, the previous secret will be overwritten.

```yaml
    objects: |
      array:
        - |
          objectPath: "/secret/foo"
          objectName: "bar"
          objectVersion: ""
        - |
          objectPath: "/secret/foo1"
          objectName: "bar"
          objectVersion: ""
```
Source: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/test/bats/tests/vault_v1alpha1_secretproviderclass.yaml

I confimed it by running `e2e-vault` of secrets-store-csi-driver.

```shell
$ k exec -it nginx-secrets-store-inline -- ls -l /mnt/secrets-store
total 4
-rw-r--r-- 1 root root 6 Jun 17 05:56 bar

$ k exec -it nginx-secrets-store-inline -- cat /mnt/secrets-store/bar
hello1
```

Ideally, It should support consul template to output in a flexible format like [vault.hashicorp.com/agent-inject-template](https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-inject-template) of Agent Sidecar Injector. but as a short-term fix, I suggest using a slash in the ObjectPath as a hyphen and a file name with a combined ObjectName to generate a unique name. For instance. `secret-foo-bar`. or should create directory based on ObjectPath (#39). I will make a PR if it is accepted.

mitsutaka:
Fixes #44 

It is inspired by https://github.com/Azure/secrets-store-csi-driver-provider-azure/pull/39, define objectAlias for user who wants to specify a filename. In the current provider behavior, objectName is used as a filename, so if you want to use the same objectName in different kv paths, you can avoid overwriting by adding objectAlias.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@ Vault and use the Secrets Store CSI driver interface to mount them into Kubernet
 
 **This is an experimental project. This project isn't production ready.**
 
-## Attribution	
-This project is forked from and initially developed by our awesome partners at Microsoft (https://github.com/deislabs/secrets-store-csi-driver). Thank you to [Rita](https://github.com/deislabs/secrets-store-csi-driver/commits?author=ritazh) and [Mishra](https://github.com/deislabs/secrets-store-csi-driver/commits?author=anubhavmishra) for pushing this great project forward.	
+## Attribution
+This project is forked from and initially developed by our awesome partners at Microsoft (https://github.com/deislabs/secrets-store-csi-driver). Thank you to [Rita](https://github.com/deislabs/secrets-store-csi-driver/commits?author=ritazh) and [Mishra](https://github.com/deislabs/secrets-store-csi-driver/commits?author=anubhavmishra) for pushing this great project forward.
 
 ## Demo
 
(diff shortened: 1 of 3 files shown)


tomhjp:
Hey @mitsutaka, sorry this has sat untended for so long - I really appreciate the idea and contribution.

We wanted to start supporting more than just the kv secrets engine, so as part of #63, we've actually done something pretty similar to this PR and split out some of the overloaded meaning from `objectName`. `objectName` will still specify the file name to write to, but `secretKey` will now specify the key within the secret to read. Unfortunately it's a breaking change, but we're trying to ensure we get all of those out of the way as early as possible, and it should be more stable going forwards from the next release.

I'll close this for now, but please feel free to leave any comments you have on that PR. (Also N.B. I haven't updated the docs in the PR yet, but will be making those changes soon if you'd rather wait for that)

PR #50, opened 2020-11-04
Enforce descending directory write path

tomhjp:
I've tried to keep this fairly targeted, but also resilient to surprising paths or deployments, including Windows nodes, while also making it easy and reliable to test on any OS. Unfortunately the existing test coverage is pretty low, so it's a big task to write more of an integrated test.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -449,10 +449,24 @@ func (p *Provider) MountSecretsStoreObjectContent(ctx context.Context, attrib ma
 			return err
 		}
 		objectContent := []byte(content)
+		if err := validateFilePath(keyValueObject.ObjectName); err != nil {
+			return err
+		}
 		if err := ioutil.WriteFile(path.Join(targetPath, keyValueObject.ObjectName), objectContent, permission); err != nil {
 			return errors.Wrapf(err, "secrets-store csi driver failed to write %s at %s", keyValueObject.ObjectName, targetPath)
 		}
 		log.Infof("secrets-store csi driver wrote %s at %s", keyValueObject.ObjectName, targetPath)
+	}
+
+	return nil
+}
+
+func validateFilePath(path string) error {
+	segments := strings.Split(strings.ReplaceAll(path, `\`, "/"), "/")
+	for _, segment := range segments {
+		if segment == ".." {
+			return fmt.Errorf("ObjectName %q invalid, must not contain any .. segments", path)
+		}
 	}
 
 	return nil
--- a/provider.go
+++ b/provider.go
@@ -10,7 +10,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"path"
 	"path/filepath"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


provider.go
```diff
@@ -458,6 +461,17 @@ func (p *Provider) MountSecretsStoreObjectContent(ctx context.Context, attrib ma
 	return nil
 }
 
+func validateFilePath(path string) error {
+	segments := strings.Split(strings.ReplaceAll(path, `\`, "/"), "/")
```
calvn, line 465: Do we need to split the path by segments? Since `..` has a special meaning on paths, we should be fine if we check the string as a whole to see if that's present, similar to the check in [here](https://github.com/hashicorp/vault/blob/master/sdk/logical/storage_view.go#L100).

provider_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestValidateFilePath(t *testing.T) {
+	// Don't use filepath.Join to generate the test cases because it calls filepath.Clean
+	// which simplifies some of the test cases into less interesting paths.
+	for _, tc := range []string {
```
jasonodonnell, line 11: I wonder if environment variables (bash expansion) is also an attack vector we need to consider? I don't think so but calling it out. I'm just worried about any other bash tricks that could cause problems here.

provider.go
tomhjp: How so? If I do `mkdir foo..bar` on my machine, that creates a single folder called `foo..bar` which I can interact with normally.

However, regardless of that, I am definitely ok with a simple `string.Contains` check for `..` as that's far harder to cheat, and still seems like a very reasonable limitation.

provider_test.go
tomhjp: That's a good thought. I think you're correct that it's not a consideration, because
a) the code that actually writes the file is within this provider, so we don't have to worry about expansion beyond the confines of this code base
b) `ioutil.WriteFile` doesn't do anything funky. It goes straight to `os.Open` which uses a syscall to open the file handle, so there shouldn't be an env expansion in there.

provider.go
calvn: You're right, `..` can be part of a filename.

xntrik (approved):
lgtm

provider.go
kalafut, line 465: filepath.ToSlash might be more canonical for the replace?

tomhjp: Yeah agreed - I did use that initially, but the fact that the behaviour of that function changes depending on the OS made it slightly more difficult to reason about and write cross-platform tests for:

```go
func ToSlash(path string) string {
	if Separator == '/' {
		return path
	}
	return strings.ReplaceAll(path, string(Separator), "/")
}
```

Using `strings.ReplaceAll` directly definitely makes the checking less correct, because (surprisingly to me) Linux/Unix file names can contain backslashes. However, I also wasn't 100% confident about whether mixing Windows and Linux was possible between pods and host node volumes, so decided to plump for the simpler fail-safe option of just treating all slashes as path separators. WDYT?

calvn: Another benefit to using `strings.Replace` is that it can protect against strings such as  `"foo/.\b./bar"` which gets past `path.Join` as `"foo/../bar"` which then gets passed on to `ioutil.WriteFile` as such ([Playground example](https://play.golang.org/p/7unVBwRjhaS)).

calvn: @tomhjp An interesting case where `strings.ReplaceAll` (or `filepath.ToSlash`) won't work: https://play.golang.org/p/BWcISJtKQpq

tomhjp: Thanks @calvn, this is a really good idea to test!

After testing it out, I think this is ok, because while \b isn't printable, the character is still there, so if you actually create that directory, it will get created as `foo/.^H./bar/`, and not get interpreted as `foo/../bar`. It's slightly clearer to play around with on your own machine, but I think this also shows it: https://play.golang.org/p/6oktHKATJLH

tomhjp: One separate thing this has made me realise is the code will currently fail for any paths specifying folders because the path doesn't exist. I'll add a call to `os.MkdirAll` to remedy.

jasonodonnell: What's the use case for slashes? I don't think it matters since all secrets are forced in a certain directory.

tomhjp: As discussed offline, it's more about handling keys with slashes gracefully. Because we do a write with the object name as the relative path, any slashes will get interpreted as path separators. We can either disallow all slashes in the object name, or implement a small feature to make them work - but what I wanted to avoid was just letting the inevitable error happen without handling it (when slashes are present), and therefore return a confusing error to the user.

jasonodonnell (approved):

calvn (approved):

PR #54, opened 2020-12-03
Add gRPC support

tomhjp:
This wraps the existing code in a gRPC server in order to support gRPC communication with the driver.

Testing steps are a bit cumbersome until we get some proper integration tests/CI inside this repo (coming in another PR soon):

```bash
# Setting up local state
git clone git@github.com/hashicorp/secrets-store-csi-driver-provider-vault.git
git -C secrets-store-csi-driver-provider-vault checkout add-grpc-support
git clone git@github.com/kubernetes-sigs/secrets-store-csi-driver.git
curl https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/files/5810149/patch.txt -o test.patch
git -C secrets-store-csi-driver/ apply ../test.patch

# Building and running the tests
kind create cluster
make -C secrets-store-csi-driver e2e-container
make -C secrets-store-csi-driver-provider-vault e2e-container
make -C secrets-store-csi-driver e2e-vault
```

The releasing process needs some refinement too, as the driver repo's integration tests and the readme use `deployment/provider-vault-installer.yaml` from master, and that references a pinned version, so I don't think there's a good way to update and release without breaking those briefly. That update will also come in a separate PR though.

[patch.txt](https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/files/5810149/patch.txt)


tvoran (commented):
Just a couple small comments. I was also able to setup and run the tests per the instructions in the PR description 👍

provider.go
```diff
@@ -224,12 +270,12 @@ func (p *Provider) login(jwt string, roleName string) (string, error) {
 	return s.Auth.ClientToken, nil
 }
 
-func (p *Provider) getSecret(token string, secretPath string, secretName string, secretVersion string) (string, error) {
+func (p *Provider) getSecret(token string, secretPath string, secretName string, secretVersion string) (string, int, error) {
```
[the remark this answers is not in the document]
tomhjp: Yeah good point, done.

tomhjp:
Thanks for the quick review @tvoran. FYI I've backed out the bulk of the changes to the deployment folder, with the idea that the tag and arguments/volumes should all just get updated shortly before releasing 0.0.7, and we might not do that straight after merging this PR.

Dockerfile
```diff
@@ -8,9 +8,8 @@ WORKDIR /bin
 
 RUN apk add --no-cache bash
 
-ADD ./install.sh /bin/install_vault_provider.sh
 COPY ./_output/secrets-store-csi-driver-provider-vault_${OS}_${ARCH}_${VERSION} /bin/secrets-store-csi-driver-provider-vault
 RUN chmod a+x /bin/secrets-store-csi-driver-provider-vault
```
jasonodonnell, line 13: Not a blocker but something we need to investigate: it appears this image is running as the root user.

README.md
```diff
@@ -88,7 +97,7 @@ spec:
     objects:  |
       array:
         - |
-          objectPath: "/foo"                    # secret path in the Vault Key-Value store e.g. vault kv put secret/foo bar=hello
+          objectPath: "v1/secret/foo"           # secret path in the Vault Key-Value store e.g. vault kv put secret/foo bar=hello
```
jasonodonnell, line 100: Note for future: we'll want to add `v1` to the paths to simplify things.

test/provider-vault-installer.yaml
```diff
@@ -0,0 +1,50 @@
...
+      labels:
+        app: csi-secrets-store-provider-vault
+    spec:
+      serviceAccountName: secrets-store-csi-driver-provider-vault
+      containers:
+        - name: secrets-store-csi-driver-provider-vault
+          image: e2e/secrets-store-csi-driver-provider-vault:latest
+          imagePullPolicy: Never
+          args:
+            - --endpoint=/provider/vault.sock
```
jasonodonnell, line 29: I see above the socket is specified at `/tmp/vault.sock` but here it's `/provider`. Is this correct?

main.go
```diff
@@ -6,22 +6,24 @@ import (
...
 )
 
 var (
-	attributes = pflag.String("attributes", "", "volume attributes")
-	secrets    = pflag.String("secrets", "", "node publish ref secret")
-	targetPath = pflag.String("targetPath", "", "Target path to write data.")
-	permission = pflag.String("permission", "", "File permission")
-	debug      = pflag.Bool("debug", false, "sets log to debug level")
-	version    = pflag.Bool("version", false, "prints the version information")
+	endpoint = pflag.String("endpoint", "/tmp/vault.sock", "path to socket on which to listen for driver gRPC calls")
```
jasonodonnell, line 21: Should this default to `/provider/vault.sock`?

jasonodonnell (approved):
Minor things, LGTM

Dockerfile
tomhjp: Good call, this is fixable, but not trivial enough to tack onto this already multi-purpose PR. I had a quick go and it should be an easy fix but non-root users need to be given permission to the mounted volumes.

main.go
tomhjp: It depends how you run it - I chose `/tmp/vault.sock` to make it easier to run locally as `/tmp` will (always?) already exist, but when crafting yaml to deploy the provider, users should match up the volume mount they create and this argument. Obviously the helm chart will get those two married up correctly by default.

test/provider-vault-installer.yaml
tomhjp: Intentional - discussion in comment below.

tvoran (approved):

PR #60, opened 2021-01-27
Add e2e tests

tomhjp:
These tests are heavily inspired by the tests in the driver repo: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/test/bats/vault.bats, but hopefully improved a bit.

I've tried to approximately draw a line between the setup/teardown responsibilities of Makefile/bats tests. Makefile handles installing everything in the `csi` namespace, and bats tests handle any further configuration + anything in the `test` namespace.

I also added linting/`go mod` jobs to CircleCI in this PR to round out the whole CI pipeline.

To run locally:

```bash
kind create cluster
make e2e-setup
make e2e-test
make e2e-teardown
```


test/bats/configs/vault.yaml
```diff
@@ -0,0 +1,83 @@
+---
```
jasonodonnell, line 1: Why not just use our helm chart instead of rolling your own?

tomhjp: Good point, this was mostly due to keeping the same method as the tests these were based on. I've switched to installing via helm chart now.

test/bats/provider.bats
```diff
@@ -0,0 +1,164 @@
...
+  [[ "$result" == "hello-sync2" ]]
+
+  result=$(kubectl --namespace=test get secret foosecret -o jsonpath="{.data.pwd}" | base64 -d)
+  [[ "$result" == "hello-sync1" ]]
+
+  result=$(kubectl --namespace=test exec $POD -- printenv | grep SECRET_USERNAME | awk -F"=" '{ print $2 }' | tr -d '\r\n')
+  [[ "$result" == "hello-sync2" ]]
+
+  run kubectl get secret --namespace=test foosecret
+  [ "$status" -eq 0 ]
```
tvoran, line 92: Should these be moved above lines 85-86? Since 85-86 are using the result of `get secret foosecret`, it seems like the status code should be checked before that.

test/bats/provider.bats
```diff
@@ -0,0 +1,164 @@
...
+      result="$(kubectl --namespace=test get secret foosecret -o json | jq '.metadata.ownerReferences | length')"
+      if [[ "$result" -eq 2 ]]; then
+          break
+      fi
+      sleep 1
+  done
+  [[ "$result" -eq 2 ]]
+
+  # Wait for secret deletion in a background process.
+  kubectl --namespace=test wait --for=delete secret foosecret &
```
tvoran, line 111: Should this include a timeout arg? Unless 30s is enough (looks like that's the default).

test/bats/provider.bats
```diff
@@ -0,0 +1,164 @@
...
+  echo $1
+  for i in {0..60}; do
+      if eval "$1"; then
+          return
+      fi
+      sleep 1
+  done
+  # Fail the test.
+  [ 1 -eq 2 ]
+}
```
tvoran, line 164: I wonder if it'd be useful to put functions like this in a _helpers.bash, then load them into bats with `load _helpers`.

tvoran (commented):
Love all the tests! Everything seemed to work for me locally, just some bats suggestions.

test/bats/provider.bats
tomhjp: Good spot, I've tried to be consistently generous with timeouts to avoid flakes, so I've added 60s as with the other waits.

test/bats/provider.bats
tomhjp: 👍 done

Makefile
```diff
@@ -7,46 +7,89 @@ BUILD_DATE=$$(date +%Y-%m-%d-%H:%M)
...
+		--wait --timeout=5m \
+		--namespace=csi \
+		--set linux.image.pullPolicy="IfNotPresent" \
+		--set grpcSupportedProviders="azure;gcp;vault"
+	helm install vault https://github.com/hashicorp/vault-helm/archive/v0.9.0.tar.gz \
+		--wait --timeout=5m \
+		--namespace=csi \
+		--set injector.enabled=false \
+		--set server.dev.enabled=true \
+		--set 'server.extraEnvironmentVars.VAULT_DEV_LISTEN_ADDRESS=0.0.0.0:8200'
```
jasonodonnell, line 77: Reminder that I think this will change in 0.10.0.

jasonodonnell (approved):
LGTM

tvoran (approved):

PR #61, opened 2021-02-01
Use vault api module

tomhjp:
- Refactor `provider.go` into several smaller `internal` packages
- Use `vault`'s `api` client instead of raw http requests
- Add some unit test coverage
- Be resilient to the socket file already existing on startup, this can happen if the previous pod didn't get a chance to shut down cleanly
- Switch to `hclog`


internal/client/client.go
```diff
@@ -0,0 +1,141 @@
...
+func loadCertFolder(pool *x509.CertPool, root string) error {
+	if err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if info.IsDir() {
+			if path != root {
+				return filepath.SkipDir
+			}
```
tomhjp, line 131: This check was previously missing, meaning the function did recurse which is a bug the unit tests caught. The function comment about not recursing is now correct, and I think that's probably the better/safer behaviour.

internal/client/client.go
```diff
@@ -0,0 +1,141 @@
+package client
```
tomhjp, line 1: Most of this file is lifted from `provider.go` untouched, except `Do` which is new, and a bug fix in `loadCertFolder`.

internal/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+		if err == nil {
+			parameters.TLSConfig.VaultSkipTLSVerify = value
+		} else {
+			return Parameters{}, err
+		}
+	}
+
+	secretsYaml := params["objects"]
+	// TODO: There is an unnecessary map under objects, instead of just directly storing an array there.
+	// Deserialisation can be simplified a fair bit if we remove it.
```
tomhjp, line 114: I plan to do this in a PR in the near future, but held off for now given the purpose and size of this PR.

internal/provider/provider.go
```diff
@@ -0,0 +1,253 @@
+package provider
```
tomhjp, line 1: If you'd like to review this file with more incremental changes, they're reasonably well broken down by commit:

* The first commit in this PR lifts code from the old `provider.go` file without much change except to pull bits out into other packages.
* The second commit switches to using the Vault API package instead of a raw HTTP client
* The third commit ensures that `ctx` values do not go ignored in those Vault API calls

internal/provider/provider.go
```diff
@@ -0,0 +1,253 @@
...
+
+func NewProvider(logger hclog.Logger) *provider {
+	p := &provider{
+		logger: logger,
+	}
+
+	return p
+}
+
+func (p *provider) getMountInfo(ctx context.Context, client *api.Client, mountName string) (string, string, error) {
```
jasonodonnell, line 44: Do we need this for anything else besides checking if a secret mount is KV v2? If not we can tell if it's KV v2 by looking at the payload. If it contains `.data.data`, then it's KV v2. If we did this instead that's one less permission CSI will need.

internal/provider/provider.go
```diff
@@ -0,0 +1,253 @@
...
+		err = writeSecret(p.logger, cfg.TargetPath, secret.ObjectName, content, cfg.FilePermission)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return versions, nil
+}
+
+func writeSecret(logger hclog.Logger, directory string, file string, content string, permission os.FileMode) error {
```
jasonodonnell, line 183: Perhaps we should turn this into a method and have a secret struct? In the future there could be more things we need to add such as leases, templates, etc. Thoughts?

internal/client/testdata/bad_directory/not-a-ca.pem
```diff
@@ -0,0 +1 @@
+Hello
```
jasonodonnell, line 1: 👋

main.go
```diff
@@ -1,152 +1,105 @@
...
+	logger.Info(fmt.Sprintf("Listening on %s", *endpoint))
 
-	s := &ProviderServer{}
+	s := &providerserver.ProviderServer{
+		Logger: logger,
+	}
 	pb.RegisterCSIDriverProviderServer(server, s)
 
 	err = server.Serve(listener)
 	if err != nil {
```
jasonodonnell, line 76: Is it worth also sending a warning to the logs?

main.go
```diff
@@ -1,152 +1,105 @@
...
-		log.Fatal(err)
+		return fmt.Errorf("error running gRPC server: %w", err)
```
jasonodonnell, line 77: Should we also warn using the logger here?

main.go
```diff
@@ -1,152 +1,105 @@
...
+	// Because the unix socket is created in a host volume (i.e. persistent
+	// storage), it can persist from previous runs if the pod was not terminated
+	// cleanly. Check if we need to clean up before creating a listener.
+	_, err := os.Stat(*endpoint)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("failed to check for existence of unix socket: %w", err)
+	} else if err == nil {
+		logger.Info("Cleaning up pre-existing file at unix socket location", "endpoint", *endpoint)
+		err = os.Remove(*endpoint)
+		if err != nil {
```
jasonodonnell, line 93: Should we also warn using the logger here?

main.go
```diff
@@ -1,152 +1,105 @@
...
-	log.SetLevel(log.InfoLevel)
-	if *debug {
-		log.SetLevel(log.DebugLevel)
+func listen(logger hclog.Logger) (net.Listener, error) {
+	// Because the unix socket is created in a host volume (i.e. persistent
+	// storage), it can persist from previous runs if the pod was not terminated
+	// cleanly. Check if we need to clean up before creating a listener.
+	_, err := os.Stat(*endpoint)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("failed to check for existence of unix socket: %w", err)
```
jasonodonnell, line 89: Should we also warn using the logger here?

internal/provider/provider.go
tomhjp: Yeah that's correct. I'll strip this out when adding generic secret support.

main.go
tomhjp: `main()` will log this as an error-level log before the provider exits with exit code 1.

main.go
tomhjp: Same as above, `main()` will log this as an error-level log before the provider exits with exit code 1.

main.go
tomhjp: Same as above, `main()` will log this as an error-level log before the provider exits with exit code 1.

main.go
```diff
@@ -1,152 +1,105 @@
...
-		log.Fatalf("Failed to listen on unix socket at %s: %v", *endpoint, err)
+		return err
 	}
 	defer listener.Close()
-	log.Infof("Listening on %s", *endpoint)
+	logger.Info(fmt.Sprintf("Listening on %s", *endpoint))
 
-	s := &ProviderServer{}
+	s := &providerserver.ProviderServer{
+		Logger: logger,
```
tvoran, line 71: Something that I've seen be useful elsewhere is to pass a named logger to different components so it's easy to tell where the logs are coming from, i.e.

```suggestion
		Logger: logger.Named("server"),
```

(Not sure if that's particularly useful here, just wanted to mention it.)

tomhjp: Yeah I considered using this. Initially I decided against it because the code base is so small it didn't seem likely that it would ever be unclear which component logs have come from. However, it's so cheap to do and I think it probably will make reading the logs easier, so I'll pepper in some usages.

internal/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+	if c.TargetPath == "" {
+		return errors.New("missing target path field")
+	}
+	if c.Parameters.VaultRoleName == "" {
+		return errors.Errorf("missing 'roleName' in SecretProviderClass definition")
+	}
+	certificatesConfigured := c.Parameters.TLSConfig.CertificatesConfigured()
+	if c.Parameters.TLSConfig.VaultSkipTLSVerify && certificatesConfigured {
+		return errors.New("both vaultSkipTLSVerify and TLS configuration are set")
+	}
```
jasonodonnell, line 159: I think there are useful dev use cases where you might not want to reject this configuration. When attempting to debug TLS CA verification issues, it can be helpful to skip verify, especially in dev environments.

internal/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+
+type TLSConfig struct {
+	VaultCAPEM         string
+	VaultCACertPath    string
+	VaultCADirectory   string
+	VaultTLSServerName string
+	VaultSkipTLSVerify bool
+}
+
+func (c TLSConfig) CertificatesConfigured() bool {
```
jasonodonnell, line 51: Future enhancement should also include a configurable to specify the SNI of the server you're connecting to.

Cert SANs can be hard to get right, so sometimes you need to override the CA verification process and say "if this SAN is present in the cert, then its valid". We have this config in Vault Agent and its very useful: https://www.vaultproject.io/docs/commands#vault_tls_server_name)

internal/version/version.go
```diff
@@ -19,7 +21,7 @@ type providerVersion struct {
 	MinDriverVersion string `json:"minDriverVersion"`
 }
 
-func getVersion() (string, error) {
+func GetVersion() (string, error) {
```
jasonodonnell, line 24: Future enhancement: should also include the version of Golang used to build it.

main.go
```diff
@@ -1,152 +1,105 @@
...
-type LogHook struct {
-	Writer    io.Writer
-	Loglevels []log.Level
+func main() {
+	logger := hclog.Default()
+	err := realMain(logger)
+	if err != nil {
+		logger.Error("Error running provider", "err", err)
+		os.Exit(1)
+	}
```
jasonodonnell, line 30: nit: HashiCorp projects tend to use the Mitchell's CLI package. This is helpful to process parameters/flags and in general have consistency across the projects. This isn't a huge deal since its exclusively running in a Docker image, but I'd suggest taking a look at Vault K8s: https://github.com/hashicorp/vault-k8s/blob/master/main.go#L12-L21.

jasonodonnell (approved):
Minor stuff but LGTM!

internal/config/config.go
tomhjp: I think this is still the correct thing to do in production, and in dev use cases we can edit the code to work around issues, so personally I'd prefer to leave this as-is

internal/config/config.go
tomhjp: SGTM, I'll add that in a future PR.

main.go
tomhjp: Will also add this in a future PR 👍

internal/provider/provider.go
tomhjp: I've given this some thought and played around a little with the code. I think you're right that something along these lines will be useful in the future, but I don't have a firm enough idea of how it will look to do it yet, so I'll defer for now.

tvoran (approved):

PR #62, opened 2021-02-04
Add go build version to version info

tomhjp:
Following up on a comment from #61.

```bash
./_output/secrets-store-csi-driver-provider-vault_darwin_amd64_0.0.7 --version
{"version":"0.0.7","buildDate":"2021-02-04-15:38","goVersion":"go version go1.15.7 darwin/amd64","minDriverVersion":"v0.0.17"}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,12 @@ IMAGE_VERSION?=$(shell git tag | tail -1)
 IMAGE_TAG=$(REGISTRY_NAME)/$(IMAGE_NAME):$(IMAGE_VERSION)
 IMAGE_TAG_LATEST=$(REGISTRY_NAME)/$(IMAGE_NAME):latest
 BUILD_DATE=$$(date +%Y-%m-%d-%H:%M)
-LDFLAGS?="-X github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.BuildVersion=$(IMAGE_VERSION) -X github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.BuildDate=$(BUILD_DATE) -extldflags "-static""
+LDFLAGS?="-X 'github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.BuildVersion=$(IMAGE_VERSION)' \
+	-X 'github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.BuildDate=$(BUILD_DATE)' \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -4,9 +4,12 @@ IMAGE_VERSION?=$(shell git tag | tail -1)
 IMAGE_TAG=$(REGISTRY_NAME)/$(IMAGE_NAME):$(IMAGE_VERSION)
 IMAGE_TAG_LATEST=$(REGISTRY_NAME)/$(IMAGE_NAME):latest
 BUILD_DATE=$$(date +%Y-%m-%d-%H:%M)
-LDFLAGS?="-X github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.BuildVersion=$(IMAGE_VERSION) -X github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.BuildDate=$(BUILD_DATE) -extldflags "-static""
+LDFLAGS?="-X 'github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.BuildVersion=$(IMAGE_VERSION)' \
+	-X 'github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.BuildDate=$(BUILD_DATE)' \
+	-X 'github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/version.GoVersion=$(shell go version)' \
```
catsby, line 9: Will this work on windows (the `shell` command)? Would it be better to set this version as a variable like we do for `BUILD_DATE`? Granted I see now that we do `shell` for `IMAGE_VERSION` as well

catsby (approved):
👍

tvoran (approved):

jasonodonnell (approved):
LGTM

Makefile
tomhjp: Thanks for pointing this out, I hadn't really considered that. I don't have a Windows machine to test on, but I think it may work? As `shell` is just a Makefile function that says "use the default shell" and I believe it will use `cmd.exe` and I _think_ `go version` will work? There are a lot of guesses there, but as mentioned it doesn't change compatibility and Windows support is a fair way off for now, so I'll leave this one as-is for this PR.

[Makefile shell docs for reference](https://www.gnu.org/software/make/manual/html_node/Shell-Function.html)
[Choosing the shell docs](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html) which make it sound way more complicated than I said above 😱

PR #63, opened 2021-02-09
Support all secret types

tomhjp:
Adds support for all secret engines via updated config options in SecretProviderClass e.g.:

```yaml
    objects: |
      - objectName: "certs"
        secretPath: "pki/issue/example-dot-com"
        secretKey: "certificate"
        secretArgs:
          common_name: "test.example.com"
          ttl: "24h"
        method: "PUT"
```

These config updates make it a breaking change for existing SecretProviderClasses, and is one of two breaking changes that are planned for the next release - the other being that we will auth to Vault as the requesting pod. Things should then be more stable after that.

As part of the updated config/support for all secret engines, this PR:
* Adds a cache that lasts for the lifetime of one mount request (i.e. one whole SecretProviderClass) to ensure repeated dynamic secret paths are not read multiple times and the username/password match
* Supports dumping the whole JSON response to disk instead of a specific `secretKey`

Other small additions:
* gRPC interceptor logging so errors always show up in the provider logs, giving one convenient location to debug failures for a specific pod.
* Use `hashicorp.jfrog.io/docker` images where possible
* Upgrade `vault-helm` chart to `v0.9.1` to pull in dev mode fix

You may want to review the `provider.bats` diff with whitespace off, as I corrected some inconsistent indentation there.


Makefile
```diff
@@ -69,12 +69,12 @@ e2e-setup: e2e-container
 		--namespace=csi \
 		--set linux.image.pullPolicy="IfNotPresent" \
 		--set grpcSupportedProviders="azure;gcp;vault"
-	helm install vault https://github.com/hashicorp/vault-helm/archive/v0.9.0.tar.gz \
+	helm install vault https://github.com/hashicorp/vault-helm/archive/v0.9.1.tar.gz \
```
jasonodonnell, line 72: This should use the helm repo instead, right?

internal/config/testdata/example-parameters-string.txt
```diff
@@ -1 +1 @@
-{"csi.storage.k8s.io/pod.name":"nginx-secrets-store-inline","csi.storage.k8s.io/pod.namespace":"test","csi.storage.k8s.io/pod.uid":"9aeb260f-d64a-426c-9872-95b6bab37e00","csi.storage.k8s.io/serviceAccount.name":"default","objects":"array:\n  - |\n    objectPath: \"v1/secret/foo1\"\n    objectName: \"bar1\"\n    objectVersion: \"\"\n  - |\n    objectPath: \"v1/secret/foo2\"\n    objectName: \"bar2\"\n    objectVersion: \"\"\n","roleName":"example-role","vaultAddress":"http://vault:8200","vaultSkipTLSVerify":"true"}
\ No newline at end of file
+{"csi.storage.k8s.io/pod.name":"nginx-secrets-store-inline","csi.storage.k8s.io/pod.namespace":"test","csi.storage.k8s.io/pod.uid":"9aeb260f-d64a-426c-9872-95b6bab37e00","csi.storage.k8s.io/serviceAccount.name":"default","objects":"- secretPath: \"v1/secret/foo1\"\n  objectName: \"bar1\"\n  method: \"GET\"\n- secretPath: \"v1/secret/foo2\"\n  objectName: \"bar2\"","roleName":"example-role","vaultAddress":"http://vault:8200","vaultSkipTLSVerify":"true"}
```
jasonodonnell, line 1: Will prettifying this file make a difference? Would be nice if it was a bit more readable.

tomhjp: The original intention was to keep it as close as possible to the real input, but prettifying doesn't break anything, and we have the integration tests for the real thing, so good idea 👍

internal/provider/provider.go
```diff
@@ -109,59 +71,101 @@ func (p *provider) login(ctx context.Context, client *api.Client, vaultKubernete
...
 	}
-	secretPrefix := s[1]
-	secretSuffix := s[2]
+}
 
-	secretMountType, secretMountVersion, err := p.getMountInfo(ctx, client, secretPrefix)
-	if err != nil {
-		return "", err
+func generateRequest(client *api.Client, secret config.Secret) (*api.Request, error) {
+	secretPath := ensureV1Prefix(secret.SecretPath)
```
jasonodonnell, line 88: I wonder if splitting on `/` would be better and checking what the first value is. Then we can prepend the API version cleanly.

test/bats/provider.bats
```diff
@@ -63,92 +78,141 @@ teardown(){
...
+
+@test "6 Dynamic secrets engine, endpoint is called only once per SecretProviderClass" {
+    # Setup postgres
+    POSTGRES_PASSWORD=$(openssl rand -base64 30)
+    kubectl --namespace=test create secret generic postgres-root \
+        --from-literal=POSTGRES_USER="root" \
+        --from-literal=POSTGRES_PASSWORD="${POSTGRES_PASSWORD}"
+    kubectl --namespace=test apply -f $CONFIGS/postgres.yaml
+    kubectl wait --namespace=test --for=condition=Ready --timeout=5m pod postgres
+    POSTGRES_POD_IP=$(kubectl --namespace=test get pod postgres -o json | jq -r '.status.podIP')
```
jasonodonnell, line 192: This could be simplified by connecting to the PostgreSQL service instead. Pod IPs are usually not used.

internal/provider/provider.go
```diff
@@ -109,59 +71,101 @@ func (p *provider) login(ctx context.Context, client *api.Client, vaultKubernete
...
+		if err != nil {
+			return "", fmt.Errorf("couldn't read secret %q: %w", secretConfig.ObjectName, err)
+		}
+		if secret == nil || secret.Data == nil {
+			return "", fmt.Errorf("empty response from %q, warnings: %v", req.URL.Path, secret.Warnings)
+		}
+
+		p.cache[key] = secret
+	} else {
+		p.logger.Debug("Secret fetched from cache", "secretConfig", secretConfig)
```
jasonodonnell, line 153: I don't think so but this secretConfig doesn't contain any secrets, right? Just method, secret path, params, etc.

test/bats/provider.bats
```diff
@@ -63,92 +78,141 @@ teardown(){
...
+    kubectl --namespace=test wait --for=condition=Ready --timeout=60s pod nginx-dynamic-creds
+
+    # Read the creds out of the pod and verify they work for a query.
+    DYNAMIC_USERNAME=$(kubectl --namespace=test exec nginx-dynamic-creds -- cat /mnt/secrets-store/dbUsername)
+    DYNAMIC_PASSWORD=$(kubectl --namespace=test exec nginx-dynamic-creds -- cat /mnt/secrets-store/dbPassword)
+    result=$(kubectl --namespace=test exec postgres -- psql postgres://${DYNAMIC_USERNAME}:${DYNAMIC_PASSWORD}@127.0.0.1:5432/db --command="SELECT usename FROM pg_catalog.pg_user" --csv | sed -n '3 p')
+
+    [[ "$result" != "" ]]
+    [[ "$result" == "${DYNAMIC_USERNAME}" ]]
 }
```
jasonodonnell, line 218: nit: can you add a test for a SecretProviderClass that mixes different secret engines? For example let's say I have an app that needs database access, certs and KV. A single secret provider class should have all the secrets my app needs in a single object. Just want to make sure we have this case covered in the tests.

test/bats/configs/vault-pki-secretproviderclass.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: secrets-store.csi.x-k8s.io/v1alpha1
+kind: SecretProviderClass
+metadata:
+  name: vault-pki
+spec:
+  provider: vault
+  parameters:
+    roleName: "example-role"
+    vaultAddress: http://vault:8200
+    vaultSkipTLSVerify: "true"
```
jasonodonnell, line 10: Can this value be omitted if the scheme is `http`? I see this in all the examples but it's not relevant to the non-tls address.

jasonodonnell (approved):
Couple small things but looks great to me!

Makefile
tomhjp: This is intentionally not adding a repo so as not to mutate state on developer's machines when they run the tests locally. I believe it is using exactly the same `.tar.gz` package files that the hashicorp helm repo would resolve to under the hood though.

test/bats/provider.bats
tomhjp: Yeah I initially looked down the route of using a service but decided it wasn't worth the extra config vs a call to find the pod IP. To go the service route, we need to turn the raw pod into a deployment, create the service with the selector labels, and then we still need to query with `kubectl` for some info about the pod, but this time for the pod's name so we can `exec` instead of for the pod's IP. Overall it seems simpler just to query the IP. WDYT?

tomhjp: Of course that's not what you'd do outside of a test, but seems ok in this very carefully curated test environment.

internal/provider/provider.go
tomhjp: Yeah that's correct, it's just config that appears under `objects` in the secret provider class.

test/bats/configs/vault-pki-secretproviderclass.yaml
tomhjp: Good point, it will currently be an error to omit this value and not have certificates configured, but we can definitely be a bit smarter about not requiring that config for `http` schemes.

test/bats/provider.bats
jasonodonnell: Just to clarify, a pod can have a service without a deployment. Services work via labels and as long as the pod matches the labels, the service will route to it.

internal/provider/provider.go
tomhjp: I had a little play with this idea but personally found the logic both harder to follow and make robust, but perhaps I don't fully understand your idea. If you think this could simplify it, do you have a little code sample?

test/bats/provider.bats
tomhjp: Good improvement 👍 done in the next PR

test/bats/provider.bats
tomhjp: Good idea, done in the next PR.

tomhjp:
Thanks for the review - as mentioned I'm deferring addressing a few of these comments into the next PR as it's already open and based on this branch.

PR #64, opened 2021-02-11
Auth as requesting pod

tomhjp:
* login now uses a freshly created service account token from the service account of the requesting pod
* token is created with a 15m ttl and bound to the lifetime of the pod
* "iss" claim in the resulting JWT will now be the value of kube-apiserver's --service-account-issuer instead of the default legacy "kubenetes/serviceaccount"
* Integration test apps (nginx) now use different service accounts depending on which secrets they will access

I will make a separate docs update PR closer to release time so that docs on `master` still make sense for the current release.


internal/provider/provider.go
```diff
@@ -47,15 +42,58 @@ type cacheKey struct {
...
+	if err != nil {
+		return "", err
+	}
+
+	ttl := int64((15 * time.Minute).Seconds())
+	resp, err := clientset.CoreV1().ServiceAccounts(podInfo.Namespace).CreateToken(ctx, podInfo.ServiceAccountName, &authenticationv1.TokenRequest{
+		Spec: authenticationv1.TokenRequestSpec{
+			ExpirationSeconds: &ttl,
+			// TODO: Support audiences as a configurable.
+			//Audiences:         []string{},
```
jasonodonnell, line 66: Can you remind me again why we're deferring this? Will this work on EKS without it?

internal/config/config.go
```diff
@@ -62,13 +74,13 @@ type Secret struct {
 	SecretArgs map[string]interface{} `yaml:"secretArgs,omitempty"`
 }
 
-func Parse(parametersStr, targetPath, permissionStr string) (Config, error) {
+func Parse(logger hclog.Logger, parametersStr, targetPath, permissionStr string) (Config, error) {
```
jasonodonnell, line 77: Wondering if we really need the logger being passed in on a parse function. I see you're logging a warning but do we think we'll need that information?

internal/provider/provider.go
```diff
@@ -47,15 +42,58 @@ type cacheKey struct {
...
+		return "", err
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return "", err
+	}
+
+	ttl := int64((15 * time.Minute).Seconds())
+	resp, err := clientset.CoreV1().ServiceAccounts(podInfo.Namespace).CreateToken(ctx, podInfo.ServiceAccountName, &authenticationv1.TokenRequest{
+		Spec: authenticationv1.TokenRequestSpec{
```
jasonodonnell, line 63: Wondering if we should add labels to reflect these were created by the CSI provider? It might be helpful if you want to query who owns them.

jasonodonnell (approved):
LGTM, minor things but overall this is awesome!

internal/provider/provider.go
```diff
@@ -47,15 +42,58 @@ type cacheKey struct {
...
 
-	req := client.NewRequest("POST", "/v1/auth/"+vaultKubernetesMountPath+"/login")
-	err := req.SetJSONBody(struct {
+	jwt, err := p.createJWTToken(ctx, params.PodInfo)
+	if err != nil {
+		return "", err
+	}
+
+	req := client.NewRequest("POST", "/v1/auth/"+params.VaultKubernetesMountPath+"/login")
+	err = req.SetJSONBody(struct {
```
kalafut, line 92: Maybe a little cleaner?

```
err = req.SetJSONBody(map[string]string{
    "role": params.VaultRoleName,
    "jwt":  jwt,
})
```

internal/provider/provider.go
mitsutaka: It is assumed that the token JWT created contains the requested Pod's namespace and service account. If so, it should be included in the Kubernetes Auth method metadata, so I think we should also include testing with the Templated policy. This proves that namespace-scoped access control is possible using the Templated policy.

mitsutaka: On the other hand, even if you prepare a ClusterRole and request a token from the API server, `MountRequest` contains the ServiceAccount Token specified by `nodePublishSecretRef`. ([source](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/provider/v1alpha1/service.proto#L49)) If you specify a secret that the pod itself can refer to in the inline volume of the pod, it can be done without requesting a token to API server. you do not need to add a new ClusterRole. What do you think?

mitsutaka (changes requested):

internal/provider/provider.go
```diff
@@ -47,15 +42,58 @@ type cacheKey struct {
...
+			BoundObjectRef: &authenticationv1.BoundObjectReference{
+				Kind:       "Pod",
+				APIVersion: "v1",
+				Name:       podInfo.Name,
+				UID:        podInfo.UID,
+			},
+		},
+	}, metav1.CreateOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to create a service account token for requesting pod %v: %w", podInfo, err)
```
calvn, line 76: This will output the entire `podInfo` object. Would `podInfo.UID` be enough context in here?

internal/provider/provider.go
```diff
@@ -47,15 +42,58 @@ type cacheKey struct {
...
+	}
+
+	p.logger.Debug("successfully generated service account token")
+	return resp.Status.Token, nil
+}
+
+func (p *provider) login(ctx context.Context, client *api.Client, params config.Parameters) (string, error) {
+	p.logger.Debug("performing vault login...")
```
calvn, line 84:
```suggestion
	p.logger.Debug("performing vault login")
```

and maybe add a success message before return on no error, similar log messages you have for `createJWTToken`.

internal/config/config.go
tomhjp: Yeah it felt a bit weird. I just wanted to try to do something to surface the deprecated configs for people who miss the release notes. I think I'll remove this after one release if that sounds ok to you.

internal/provider/provider.go
tomhjp: It would be if the pod still exists when the user reads the logs. But I left pod name + namespace in there in case the pod has since been deleted so a human reader can still understand some context. The service account name being logged is kind of just collateral but I figured does no harm.

internal/provider/provider.go
tomhjp: > Wondering if we should add labels to reflect these were created by the CSI provider?

Is this on the assumption that a k8s secret is created for the token? I just had a check, and by default it doesn't get mounted into a secret, it just lives in the API store until it expires 15m later.

> It is assumed that the token JWT created contains the requested Pod's namespace and service account. If so, it should be included in the Kubernetes Auth method metadata, so I think we should also include testing with the Templated policy. This proves that namespace-scoped access control is possible using the Templated policy.

Sorry I'm not sure if I've understood this comment correctly, but some hopefully relevant comments:
* We do explicitly bind a vault policy to a specific [service account + namespace] in the tests, but vault doesn't care which service account token you use, as long as the issuer you put in the vault k8s auth endpoint matches what it finds in the JWT
* There isn't a negative test for a service account from the wrong namespace, but I guess IMO that would feel more like testing kubernetes/vault k8s auth method than CSI.

> On the other hand, even if you prepare a ClusterRole and request a token from the API server, MountRequest contains the ServiceAccount Token specified by nodePublishSecretRef. (source) If you specify a secret that the pod itself can refer to in the inline volume of the pod, it can be done without requesting a token to API server. you do not need to add a new ClusterRole. What do you think?

Thanks for this comment, it's a very interesting idea, and not one I had considered in depth before. Some thoughts on the tradeoff of this against the PR's current approach:
* `+` No ClusterRole required, fewer privileges for the provider
* `-` Harder to deploy as you don't know a service account's k8s secret name until the SA is deployed
* `-` Assuming you use the default service account token created by kubernetes, then the JWT has no expiry
* `+/-` JWT issuer remains as the default `kubernetes/serviceaccount`, it's one less thing for vault operators to configure but it also makes the validation check less meaningful

Ultimately, I would want to replace either solution with the ability for the CSI driver to request a bound service account token as part of the `NodePublishVolume` request: [KEP link](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1855-csi-driver-service-account-token/README.md). That's behind a feature gate in alpha for 1.20, so it's very new, but it would get us the best of both worlds.

Given the tradeoffs, I'd suggest for now we open an issue that tracks utilising `nodePublishSecretRef` as a separate feature, and see how quickly the upstream improvement comes down the pipe in the meantime. What do you think?

internal/provider/provider.go
tomhjp: I've tested this out on EKS, AKS and GKE now, and it works without setting audiences as long as you set the issuer correctly when configuring vault. Most services default to returning an audience which is the same as the issuer, except EKS which returns `"https://kubernetes.default.svc"`, which I don't quite understand the motivation for.

Legacy style service account tokens (as we were using before) don't have an audience set, so this feature would be purely additive, and I'd like to do a bit more research before implementing and documenting it, but definitely one to follow up on.

internal/config/config.go
jasonodonnell: Sounds good to me.

internal/provider/provider.go
mitsutaka: > There isn't a negative test for a service account from the wrong namespace, but I guess IMO that would feel more like testing kubernetes/vault k8s auth method than CSI.

Yes, but the service account toke is created by the vault provider's request so that we should confirm that we can use entity metadata in the policy. btw, I was talking about [policy templating](https://learn.hashicorp.com/tutorials/vault/policy-templating) with vault provider + Kubernetes auth method. 

> Ultimately, I would want to replace either solution with the ability for the CSI driver to request a bound service account token as part of the NodePublishVolume request: KEP link. That's behind a feature gate in alpha for 1.20, so it's very new, but it would get us the best of both worlds.

Yes, I also follow the KEP that will be implemented secrets-store-csi-driver, not the provider itself  https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/453 :) I assume this feature is an alternative way to authenticate with requesting pod for Kubernetes that is older than 1.20(and for the probable majority of users). so that I thought it was not good to give an extra role to the provider side, even though alternative way.

> - Harder to deploy as you don't know a service account's k8s secret name until the SA is deployed
> - Assuming you use the default service account token created by kubernetes, then the JWT has no expiry
> +/- JWT issuer remains as the default kubernetes/serviceaccount, it's one less thing for vault operators to configure but it also makes the validation check less meaningful

I agree with that. but we can manage not to use the default service account by [`bound_service_account_names`](https://www.vaultproject.io/api-docs/auth/kubernetes#bound_service_account_names). I'm not sure yet HashiCorp's recommended configuration for Kubernetes auth method. Sure, the JWT doesn't have an expiration date, but the vault token also has a TTL, so I didn't think it would be a big deal. I'm still not sure what kind of case the JWT needs to expire.

> Given the tradeoffs, I'd suggest for now we open an issue that tracks utilising nodePublishSecretRef as a separate feature, and see how quickly the upstream improvement comes down the pipe in the meantime. What do you think?

Sure. I want to be clear better way for the requested pod's service account. and will see the activity on csi-driver.

tomhjp: > Yes, but the service account toke is created by the vault provider's request so that we should confirm that we can use entity metadata in the policy.

Oh I see, yeah that seems like a nice test. I'll add that.

> Assuming you use the default service account token created by kubernetes

On this point, I meant the default token of a non-default service account, not the `default` service account. e.g. if you create a service account called `nginx-sa`, kubernetes will create a k8s secret called something like `nginx-sa-token-xyzxy`, where the suffix isn't known until the service account is deployed. My assumption was that this is the k8s secret you would reference with `nodeSecretPublishRef`. So operators would have to split their deployment into two parts. One to deploy the service account and query the created token's name, and then another to deploy the pod/volume spec that references that token name.

> I'm still not sure what kind of case the JWT needs to expire.

This is just a consideration for defence in depth. When breaches happen, secrets that live forever can make them higher impact and harder to respond to, so the less we depend on them the better.

mitsutaka: We can [manually create API token using existing ServiceAccount with the fixed name](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#to-create-additional-api-tokens) so that operator doesn't need to split their deployment into two parts.

I agree that the service account token will live forever unless the user deletes it. Nevertheless, using ClusterRole with the provider feels like a bad practice. I would agree if the CSI driver will provide a short-term service account token through the `MountRequest` call. However, it is not certain if secrets-store-csi-driver needs that feature. it seems to be a good feature for users prior to v1.20. but this is a rough idea. I'm still thinking to support reading `nodePublishSecretRef` rather than cluster-wide permission as an option.

PR #66, opened 2021-02-16
Support mTLS

tomhjp:
* Deprecated `vaultCAPEM` config value - it doesn't seem like a very useful deployment mechanism for trusted CAs anyway, and was the only reason the `client` package implemented creating its own `http` client.
* Added TLS with Vault to existing integration tests.
* Added mTLS support via Vault's API package.
* Added mTLS to all but one of the integration tests' secret class providers to give test coverage with and without.


internal/client/client.go
```diff
@@ -2,38 +2,27 @@ package client
...
-	"path/filepath"
 
 	"github.com/hashicorp/secrets-store-csi-driver-provider-vault/internal/config"
 	"github.com/hashicorp/vault/api"
-	"github.com/pkg/errors"
-	"golang.org/x/net/http2"
 )
 
-func New(params config.Parameters) (*api.Client, error) {
+func New(vaultAddress string, tlsConfig config.TLSConfig) (*api.Client, error) {
```
jasonodonnell, line 11: Do we have documented somewhere the names we expect these files to have? In Helm we can setup things correctly but if a user tries to install this manually, it might be a gotcha.

jasonodonnell (approved):
Minor questions but LGTM!

internal/config/config.go
```diff
@@ -61,9 +61,10 @@ type PodInfo struct {
 }
 
 func (c TLSConfig) CertificatesConfigured() bool {
```
calvn, line 63: Is this func still needed? It looks like all calls/references are being removed.

internal/config/config.go
```diff
@@ -155,15 +160,6 @@ func (c *Config) Validate() error {
...
 		return errors.Errorf("missing 'roleName' in SecretProviderClass definition")
 	}
-	certificatesConfigured := c.Parameters.TLSConfig.CertificatesConfigured()
-	if c.Parameters.TLSConfig.VaultSkipTLSVerify && certificatesConfigured {
-		return errors.New("both vaultSkipTLSVerify and TLS configuration are set")
-	}
-	if !certificatesConfigured &&
-		!c.Parameters.TLSConfig.VaultSkipTLSVerify &&
-		strings.HasPrefix(c.Parameters.VaultAddress, "https") {
-		return errors.New("no TLS configuration and vaultSkipTLSVerify is false but vault address scheme is https")
```
calvn, line 165: Since we're not doing validation here, does it mean that we rely on the first client request failure to indicate that TLS configuration was not set up properly?

internal/client/client.go
tomhjp: The secret provider class config takes in file names, so I think they can be anything unless I'm missing something else, e.g. `vaultTLSClientCertPath: /mnt/tls/client.crt`.

internal/config/config.go
tomhjp: Good spot, thanks!

internal/config/config.go
tomhjp: We now rely on the validation logic from the Vault API package's `ConfigureTLS` function instead. Previously we didn't use that function all the time in order to continue supporting the `vaultCAPEM` option to pass the CA as a string in the SecretProviderClass. But it didn't seem worth continuing to support that given it meant maintaining an awful lot of code to roll our own HTTP client. And it also seems like a weird deployment pattern to paste your CA certificate into a secret provider class config anyway. That switch is the main reason this PR manages to net delete lines despite adding a new feature 😄

internal/client/client.go
jasonodonnell: Ah OK, I assumed `path` was just the path, like we do in Vault's configuration.

PR #71, opened 2021-03-16
Add health handler

jasonodonnell:
In order to support readiness and liveness probes, we need a health handler. This creates a simple HTTP handler that return a 204 if it's receiving traffic.

The current e2e tests will automatically test this feature because I added the probes to the deployment and the setup code uses the `kubectl --wait` functionality. If the probes aren't working the pods will never go ready.

### DIFF
--- a/main.go
+++ b/main.go
@@ -55,18 +55,6 @@ func realMain(logger hclog.Logger) error {
 		return err
 	}
 
-	// Create health handler
-	mux := http.NewServeMux()
-	ms := http.Server{
-		Addr:    *healthAddr,
-		Handler: mux,
-	}
-	defer ms.Shutdown(context.Background())
-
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -52,6 +55,18 @@ func realMain(logger hclog.Logger) error {
 		return err
 	}
 
+	// Create health handler
+	mux := http.NewServeMux()
```
calvn, line 59: Can we move this down as well so that it's closer to where it being called? It also ensures that the defer Shutdown is called first on return to signal that we're no longer healthy early on.

jasonodonnell: Done 👍

calvn (approved):

tvoran (approved):

tomhjp (approved):

PR #72, opened 2021-03-17
Add semver 2 versions to test charts

tvoran:
It looks like helm 3.5.3 is more strict about [chart versions and semver](https://helm.sh/docs/topics/charts/#charts-and-versioning). I was getting this error when setting up e2e tests locally:

```
helm install vault-bootstrap test/bats/configs/vault \
		--namespace=csi
Error: validation: chart.metadata.version "test" is invalid
make: *** [e2e-setup] Error 1
```

This PR just adds a 0 to satisfy semver 2.

### DIFF
--- a/test/bats/configs/nginx/Chart.yaml
+++ b/test/bats/configs/nginx/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v3
 name: nginx-test
-version: test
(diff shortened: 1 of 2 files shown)


tomhjp (approved):
Thanks!

PR #73, opened 2021-03-18
Make the e2e-teardown a little more robust

tvoran:
These changes were useful when cleaning up a partial setup locally, otherwise the csi namespace wasn't being deleted (and subsequent test setups would then fail).


tomhjp (approved):
LGTM 👍

PR #79, opened 2021-03-24
Update docs and deployment yaml for 0.1.0 release

tomhjp:
* Update some references to old name
* Update deployment yaml for next release

Issue #65, opened 2021-02-15
Hashicorp Vault Enterprise namespace support


wilmanpabel:
Namespace support need in Hashicorp Vault enterprise.
maybe as a parameter in secretProviderClass yaml like vaultNamespace:

<img width="856" alt="Screen Shot 2021-02-15 at 18 06 01" src="https://user-images.githubusercontent.com/8837805/108000654-786ce480-6fb8-11eb-844a-695ca08693df.png">

tomhjp:
This is a great feature request, thanks for raising it! I think adding `vaultNamespace` to the SecretProviderClass yaml would be perfect.

For now I think you should still be able to use namespaces by prefixing your secret path with the namespace (do let me know if that doesn't work), but we'll add the namespace config option as well.

PaulJDix:
This is a critical piece for us to make use of the CSI Driver for vault at our company. No additional comments to add to that above.

tomhjp:
Added in #84, we'll be releasing it later this week.

PR #80, opened 2021-03-24
Update changelog version

tomhjp:
Missed this update ready for release from #79

## 0.1.0 (March 24th, 2021)

CHANGES

* All secret engines are now supported [[GH-63](https://github.com/hashicorp/vault-csi-provider/pull/63)]
  * **This makes several breaking changes to the configuration of the SecretProviderClass' `objects` entry**
  * There is no top-level `array` entry under `objects`
  * `objectVersion` is now ignored
  * `objectPath` is renamed to `secretPath`
  * `secretKey`, `secretArgs` and `method` are newly available options
  * `objectName` no longer determines which key is read from the secret's data
  * If `secretKey` is set, that is the key from the secret's data that will be written
  * If `secretKey` is not set, the whole JSON response from Vault will be written
  * `vaultSkipTLSVerify` is no longer required to be set to `"true"` if the `vaultAddress` scheme is not `https`
* The provider will now authenticate to Vault as the requesting pod's service account [[GH-64](https://github.com/hashicorp/vault-csi-provider/pull/64)]
  * **This is likely a breaking change for existing deployments being upgraded**
  * vault-csi-provider service account now requires cluster-wide permission to create service account tokens
  * auth/kubernetes mounts in Vault will now need to bind ACL policies to the requesting pods'
    service accounts instead of the provider's service account.
  * `spec.parameters.kubernetesServiceAccountPath` is now ignored and will log a warning if set
* The provider now supports mTLS [[GH-65](https://github.com/hashicorp/vault-csi-provider/pull/65)]
  * `spec.parameters.vaultCAPem` is now ignored and will log a warning if set. **This is a breaking change**
  * `spec.parameters.vaultTLSClientCertPath` and `spec.parameters.vaultTLSClientKeyPath` are newly available options
IMPROVEMENTS
* The provider now uses the `hashicorp/vault/api` package to communicate with Vault [[GH-61](https://github.com/hashicorp/vault-csi-provider/pull/61)]
* `--version` flag will now print the version of Go used to build the provider [[GH-62](https://github.com/hashicorp/vault-csi-provider/pull/62)]
* CircleCI linting, tests and integration tests added [[GH-60](https://github.com/hashicorp/vault-csi-provider/pull/60)]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+## 0.1.0 (March 24th, 2021)
+
 CHANGES:
 
 * All secret engines are now supported [[GH-63](https://github.com/hashicorp/vault-csi-provider/pull/63)]


jasonodonnell (approved):
LGTM

Issue #88, opened 2021-04-28
return files in grpc response to csi-driver


manedurphy:
The purpose here would be to allow the csi-driver to write the secrets to the filesystem of pods instead of the provider. This behavior will be added behind a feature flag similar to how it was done gcp-provider seen here: https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/issues/98

I would like to be assigned for this task

tomhjp:
Thanks for taking a look at this! The feature flag sounds like a good plan to allow people to upgrade the driver and the provider independently. I think I'd like to make it the default behaviour eventually, but we'll give people some advanced notice of that in the first release this appears in.

PR #89, opened 2021-05-04
Feature: Send secrets to Secrets-Store-Driver ("write_secrets" bool flag)

manedurphy:
**Which issue does this PR address?**
#88 

**What does this feature/enhancement do?**
By setting the `write_secrets` flag to `false`, the provider sends secrets to the Secrets-Store-Driver instead of writing to file system.

Note: You will see that I've commented some code out where I made changes, as well as a "change here" comment in several places. I intend to remove these once I have received a review and the changes are ready for merging. I will also squash the changes into one commit.

I did want an opinion on `MountSecretsStoreObjectContentResponse` struct I made, I was wondering if there were better practices I could be following there. I was also wondering if I should upgrade the version of Go to `1.16.3`?


tomhjp (commented):
Thanks for the contribution! Generally looks pretty good to me so far, and thanks for the bug fix in there.

There is some unfortunate flakiness in the integration tests that I've had trouble reproducing outside of CI, although the failure in this PR looks systematic and likely related to the CSI driver version update. Line 182 in `provider.bats` expects two owners, and I think we may now expect 1 due to https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/493, although I haven't fully dug into that.

Feel free to self-review on GitHub to start off discussions about specific questions, now that the PR's open that may be a better workflow for you than the code comments.

Some of my comments are suggesting slightly broader improvement as well, LMK what you think.

In terms of things we should add before landing:
* Some test coverage - as discussed offline this may be a little tricky and we can sync some more about the approach
* A plan for documentation - I think we should probably add a page to https://www.vaultproject.io/docs/platform/k8s/csi/ for command line flags at this stage, I can help you sort that with a PR on the hashicorp/vault repo.
* A changelog note

manedurphy:
Added some of the recommended changes, as well as tried to isolate the problem from the breaking test (test 2 in bats)

Changes:
1. Removal of `handleMountRequest` in `server.go`. The response in `MountSecretsStoreObjectContent` now has return of type `*pb.MountResponse`
2. In order for test 2 in `provider.bats` to pass, the length of the `ownerReferences` array had to be changed to 1. At no point did it have more than 1 owner reference, it just had the ReplicaSet from the nginx deployment. I am wondering other object is expected to act as an owner to this secret?

Appreciate the feedback

tomhjp (commented):
Thanks for the updates, the code is looking good, just a few minor follow up suggestions. I think the biggest thing left to address is tests. I need to give that some more thought on how best to cover this feature while not abandoning the old code path, but I'm open to suggestions.

tomhjp (approved):
LGTM, thanks for the contribution! I've successfully run the tests with this feature enabled, and I'll follow up separately with another PR to automate that test coverage.

Issue #82, opened 2021-03-24
Configure Vault connection details in Provider pod


tomhjp:
Currently, Vault connection details are specified in the SecretProviderClass, e.g.:

```yaml
apiVersion: secrets-store.csi.x-k8s.io/v1alpha1
kind: SecretProviderClass
metadata:
  name: vault-kv
spec:
  provider: vault
  parameters:
    roleName: "kv-role"
    vaultAddress: https://vault:8200
    vaultCACertPath: /mnt/tls/ca.crt
    vaultTLSClientCertPath: /mnt/tls/client.crt
    vaultTLSClientKeyPath: /mnt/tls/client.key
    objects: |
      - objectName: "secret-1"
        secretPath: "secret/data/kv1"
        secretKey: "bar1"
```

However, this awkwardly punctures the separation of concerns between the application namespace and the CSI provider's namespace. The application-namespaced SPC has to know about the contents of the provider pod's file system in order to configure which TLS certificates to use. The argument for `vaultAddress` being out of place is less clear, but would also arguably make more sense to be configured on the Vault provider side rather than on the application side.

If we move Vault connection details into the provider pod's configuration, in addition to a better separation of concerns, it will also allow us to safely deploy Vault Agent as a side car for the provider. That would then give us lots of nice features like caching and automatic lease renewal almost for free.

isugimpy:
In this vein, the defaultVaultAddress and defaultVaultKubernetesMountPath are hardcoded constants.  By changing both to be configurable at the Provider level, a sane default could be set for the Provider to use if the SecretProviderClass does not set them.  This would be a helpful abstraction for users, and allow them to still override if they want to.  Functionally, all that would be changing here is adding a configuration value with defaults for these two variables, instead of making them consts.

ambis:
I would love to see the client tls cert configs to be moved to the provider config as well.

The client TLS cert configs (eg. vaultTLSClientKeyPath) are tightly tied to the vault instance (ie. defaultVaultAddress), and having to define them on application level feels just wrong, especially when the vault address has been defaulted.

PR #96, opened 2021-05-15
Feature:  Allow changing default Vault address and Kubernetes mount

isugimpy:
This PR is a step toward #82 in that it allows setting the Vault address and Kubernetes mount globally, instead of them being set as constants in the code.  It does not remove the ability for users to specify vaultAddress or vaultKubernetesMountPath on their SecretProviderClass.  This is a big advantage because it allows the cluster operator to set sane defaults for a given cluster that users won't have to be aware of to use the CSI.


isugimpy:
Oh, and just to clarify since I hadn't said so explicitly:  I've tested this on a live cluster and everything seems to work as expected.

main.go
```diff
@@ -24,6 +24,8 @@ var (
 	debug       = flag.Bool("debug", false, "sets log to debug level")
 	healthAddr  = flag.String("health_addr", ":8080", "configure http listener for reporting health")
 	selfVersion = flag.Bool("version", false, "prints the version information")
+	vaultAddr   = flag.String("vault_addr", "https://127.0.0.1:8200", "default address for connecting to Vault")
```
tomhjp, line 27: Now that I look at this closely, `--health_addr` should really be `--health-addr` to be consistent with the rest of our tools. Please could you switch to dashes? And I'll make changes separately to fix `health_addr`.

tomhjp (commented):
Thanks for raising this! It all looks good to me, it's just raised one existing issue with the flags that I'd like to fix before we add any more with underscores.

Please could you also add a changelog entry? Other than that, I think the test coverage looks good so this is pretty much ready to land.

main.go
isugimpy: Done!

isugimpy:
Fixed things up in the requested way.  If possible, after this is merged I'd greatly appreciate y'all cutting a release for this.  We're wanting to start using this in production, and it'd be super helpful to not have to run a forked binary instead of the official release.

tomhjp (approved):
Thank you! It shouldn't be long to wait until the next release

tomhjp:
Apologies I butchered the resolve conflicts step a bit after merging the other open PR, hopefully can get a set of green CI checks and merge this shortly.

isugimpy:
No worries, greatly appreciate the review and approval!

PR #100, opened 2021-06-02
Consistent flags and CI for --write-secrets=false

tomhjp:
This PR makes our flags consistent on using dashes instead of underscores as with our other products. To help support that in a forwards-compatible way, I've left `--health_addr` in there for now (because it's the only underscore argument we've released) but marked it as deprecated, and if someone specifies both `--health_addr` and `--health-addr`, the one that occurs later in the command line wins.

This PR also adds a couple of `make` targets and updated CI config to support running tests with both behaviours of `--write-secrets`. It's a pretty significant change in the operation of the provider because it controls whether the provider or the driver writes the secrets, so I'd like to keep good test coverage of both flows until `--write-secrets=true` becomes a deprecated behaviour a few versions down the line.

Both of these entries will get an advanced notice in our next release changelog explaining that they will be a breaking change in a future release, but both can also be fixed by users well ahead of time once the next release is out.

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,16 +19,6 @@ import (
 	pb "sigs.k8s.io/secrets-store-csi-driver/provider/v1alpha1"
 )
 
-var (
-	endpoint     = flag.String("endpoint", "/tmp/vault.sock", "path to socket on which to listen for driver gRPC calls")
-	debug        = flag.Bool("debug", false, "sets log to debug level")
-	healthAddr   = flag.String("health_addr", ":8080", "configure http listener for reporting health")
-	selfVersion  = flag.Bool("version", false, "prints the version information")
-	vaultAddr    = flag.String("vault-addr", "https://127.0.0.1:8200", "default address for connecting to Vault")
-	vaultMount   = flag.String("vault-mount", "kubernetes", "default Vault mount path for Kubernetes authentication")
-	writeSecrets = flag.Bool("write_secrets", true, "write secrets directly to filesystem (true), or send secrets to CSI driver in gRPC response (false)")
-)
-
 func main() {
 	logger := hclog.Default()
 	err := realMain(logger)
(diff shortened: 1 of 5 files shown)


tomhjp:
I've put a note in the changelog slating `--write-secrets=false` as the default in 2 major versions from now, as the driver is looking for providers to support it as the default by v0.0.23 which is roughly a month away. See https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/551 for context.

main.go
```diff
@@ -39,6 +29,17 @@ func main() {
...
+	var (
+		endpoint     = flag.String("endpoint", "/tmp/vault.sock", "path to socket on which to listen for driver gRPC calls")
+		debug        = flag.Bool("debug", false, "sets log to debug level")
+		selfVersion  = flag.Bool("version", false, "prints the version information")
+		vaultAddr    = flag.String("vault-addr", "https://127.0.0.1:8200", "default address for connecting to Vault")
+		vaultMount   = flag.String("vault-mount", "kubernetes", "default Vault mount path for Kubernetes authentication")
+		writeSecrets = flag.Bool("write-secrets", true, "write secrets directly to filesystem (true), or send secrets to CSI driver in gRPC response (false)")
+		healthAddr   = new(string)
+	)
+	flag.StringVar(healthAddr, "health_addr", "", "deprecated, please use --health-addr")
```
jasonodonnell, line 41: Should this still have a default of `:8080`?

jasonodonnell (approved):
LGTM

tvoran (approved):

main.go
tomhjp: I tested this and `flag.StringVar` immediately sets the contents of the pointer you pass it, so only the second call's default has an effect. Otherwise, the default just turns up in the usage help text and we only want usage to be the deprecated notice, so I've omitted it. This is 100% gross though and is why I've set a timeline for actually removing the flag value rather than leaving it around forever.

Issue #115, opened 2021-10-15
fetching complex JSON from Vault


rkennedy-tpl:
from the changelog...

```
If secretKey is set, that is the key from the secret's data that will be written
If secretKey is not set, the whole JSON response from Vault will be written
```

I have complex JSON data stored in Vault, example (heavily redacted, use your imagination):

![image](https://user-images.githubusercontent.com/76958947/137504481-e3a7c0a7-261d-4c51-9b91-3b6f0bb1fabf.png)

the data shows up correctly when `secretKey` is excluded. however, the "whole JSON response" shows up and it includes extra details, ex:

```
$ head /mnt/secrets-store/fabricProfile 
{"request_id":"deadbeef-xxxx-yyyy-zzzz-moreredactedhex","lease_id":"","lease_duration":0,"renewable":false,"data":{"data":{"certificateAuthorities":{"hlf-ca--unicorn":{....
```

I tried setting `secretKey` to a variety of things like `data` `Data` `.data` to see if I could trick the provider into sending me the JSON body without the extraneous response "envelope".

is there a specific key I can use when fetching JSON blocks? I was really hoping to leverage JSON instead of only key/value pairs. or is this something that would require code-changes to the `vault-csi-provider` to support?

agates4:
Got it working

In the secret provider, no `secretKey` should be specified
```
      - objectName: "vault-gcr-json-key-object"
        secretPath: "secret/data/buildkite/gcr_json_key"
```

as a mounted secret, `.Data.data` should be specified as the key
```
    - secretName: vault-gcr-json-key
      type: Opaque
      data: 
        - objectName: vault-gcr-json-key-object
          key: .Data.data
```

cheers 🥂

rkennedy-tpl:
I mean, `vault-csi` is still writing a file today, it's just "enveloped" with some extraneous metadata. I might tell my developers to "expect to de-envelope configs" or I can throw some random shim-code in somewhere that de-envelopes the configs before starting the application workloads. this is by no means a "game-breaking bug", it's just "mildly inconvenient" for some SDE Teams, who I have little control over ;)

[...]

PR #121, opened 2021-12-03
Feature: Allow retrieval of a JSON from secretKey value

agates4:
https://github.com/hashicorp/vault-csi-provider/issues/115
This issue describes how if a key with the value of a JSON is specified, it will not be successfully returned because it is not a string. However, if no `secretKey` is specified, the entire JSON response is returned, with no option for selecting a specific inner JSON. This means that the standard `data: {}, metadata: {}` response of Vault is at the top level of the JSON.

**Problem Use-case**
For a sample use-case, this means that a JSON service account secret cannot be successfully mounted into a Kubernetes pod using the vault-csi-provider, without the additional properties of the fully returned response.

**Solution**
Allow the retrieval of a JSON from the `secretKey` value.

**Updates**
- If the value of `secretKey` within the secret data is not a string,
  - Check if it is a valid JSON and return the full JSON if it is
- Test for the return of an expected JSON

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -166,9 +166,16 @@ func keyFromData(rootData map[string]interface{}, secretKey string) (string, err
 
 	content, ok := data[secretKey].(string)
 	if !ok {
+		bytes, err := json.Marshal(data[secretKey])
+		if err == nil {
+			// Only return a valid JSON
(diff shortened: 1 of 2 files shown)


agates4:
making new PR with my signed commit and verified email in config so two users aren't specified for the CLA.

PR #122, opened 2021-12-03
Feature: Allow retrieval of a JSON from secretKey value

agates4:
https://github.com/hashicorp/vault-csi-provider/issues/115
This issue describes how if a key with the value of a JSON is specified, it will not be successfully returned because it is not a string. However, if no `secretKey` is specified, the entire JSON response is returned, with no option for selecting a specific inner JSON. This means that the standard `data: {}, metadata: {}` response of Vault is at the top level of the JSON.

**Problem Use-case**
For a sample use-case, this means that a JSON service account secret cannot be successfully mounted into a Kubernetes pod using the vault-csi-provider, without the additional properties of the fully returned response.

**Solution**
Allow the retrieval of a JSON from the `secretKey` value.

**Updates**
- If the value of `secretKey` within the secret data is not a string,
  - Check if it is a valid JSON and return the full JSON if it is
- Test for the return of an expected JSON

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -166,6 +166,13 @@ func keyFromData(rootData map[string]interface{}, secretKey string) (string, err
 
 	content, ok := data[secretKey].(string)
 	if !ok {
+		bytes, err := json.Marshal(data[secretKey])
+		if err == nil {
+			// Return a valid JSON, preventing return of other types as string
+			if strings.Contains(string(bytes), "{") {
+				return string(bytes), nil
+			}
+		}
 		return "", fmt.Errorf("failed to get secret content %q as string", secretKey)
 	}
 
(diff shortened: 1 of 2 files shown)


agates4:
Requesting a review from @tomhjp @tvoran

agates4:
Hey folks it's been 1 month since submission @tomhjp @tvoran.

Any feedback is better than no feedback.

tomhjp:
Hi @agates4, thanks for raising this PR and UX issue. I think you should be able to mount JSON content with today's code, but you would have to escape it all into a big long string which is obviously far from ideal.

I needed to play around with this myself to properly understand the way we should process string vs JSON data and whether we needed any special casing. In the process I've expanded the change set a bit to have as little special casing as possible and also pass around `[]byte` contents instead of `string` as that now seems to make more sense. I've opened #126, which should address this. Feel free to leave any comments there.

PR #126, opened 2022-01-10
Support JSON values for secretKey

### LINKED ISSUE #115: fetching complex JSON from Vault

tomhjp:
Resolves #115 and based on code from #122. You can put arbitrary JSON data into `kv`'s values (and theoretically other secret engines could do the same) using a JSON file and a command like `vault kv put secret/test @test.json`, and this PR allows `vault-csi-provider` to specifically select that JSON data without having to escape your secret JSON into a big ugly string when storing it in Vault.

This PR does _not_ address arbitrary paths within the JSON though. Vault's API itself enforces that the top-level data object is a map, and users should be able to select any single top-level key from that object.

Strings remain special-cased so that we continue to write string values without wrapping them in quotes.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -152,7 +152,7 @@ func generateRequest(client *api.Client, secret config.Secret) (*api.Request, er
 	return req, nil
 }
 
-func keyFromData(rootData map[string]interface{}, secretKey string) (string, error) {
+func keyFromData(rootData map[string]interface{}, secretKey string) ([]byte, error) {
 	// Automatically parse through to embedded .data.data map if it's present
 	// and the correct type (e.g. for kv v2).
 	var data map[string]interface{}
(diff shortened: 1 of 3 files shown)


agates4:
Hey Tom, this looks great to me. Thanks for the thorough changes and response!

internal/provider/provider.go
```diff
@@ -164,35 +164,48 @@ func keyFromData(rootData map[string]interface{}, secretKey string) (string, err
...
+		return []byte(content), nil
+	}
+
+	// Arbitrary data can be returned in the data field of an API secret struct.
+	// It's already been Unmarshalled from the response, so in theory,
+	// marshalling should never realistically fail, but don't log the error just
+	// in case, as it could contain secret contents if it does somehow fail.
+	if content, err := json.Marshal(data[secretKey]); err == nil {
+		return content, nil
 	}
```
tvoran, line 179: Interesting, I naively thought these would be flipped (checking for json and then string), but of course data is already unmarshalled here.

internal/provider/provider.go
```diff
@@ -164,35 +164,48 @@ func keyFromData(rootData map[string]interface{}, secretKey string) (string, err
...
-			return "", fmt.Errorf("couldn't read secret %q: %w", secretConfig.ObjectName, err)
+			return nil, fmt.Errorf("couldn't read secret %q: %w", secretConfig.ObjectName, err)
 		}
 		if secret == nil || secret.Data == nil {
-			return "", fmt.Errorf("empty response from %q, warnings: %v", req.URL.Path, secret.Warnings)
+			return nil, fmt.Errorf("empty response from %q, warnings: %v", req.URL.Path, secret.Warnings)
+		}
+
+		for _, w := range secret.Warnings {
+			p.logger.Warn("warning in response from Vault API", "warning", w)
```
tvoran, line 208: Nice addition!

tvoran (approved):

internal/provider/provider.go
tomhjp: Yeah it took me a surprising amount of iteration to arrive at this. I wanted to just have JSON, but then the extra quotes ruin the most common usage.

PR #128, opened 2022-01-10
Use CRT for releases

tomhjp:
Updates to use CRT for releases:

* New GitHub action to create the artefacts for promotion
* New CRT config files
* Dockerfile restructured for dev/CI targets
* Various Makefile updates

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,12 +61,7 @@ jobs:
           GOARCH: ${{ matrix.arch }}
         run: |
           mkdir dist out
-          CGO_ENABLED=0 go build \
-            -ldflags "-X github.com/hashicorp/vault-csi-provider/internal/version.BuildVersion=$(make version) \
-              -X github.com/hashicorp/vault-csi-provider/internal/version.BuildDate=$(date +%Y-%m-%d-%H:%M) \
-              -X github.com/hashicorp/vault-csi-provider/internal/version.GoVersion=$(go version | cut -d' ' -f3)" \
-            -o dist/ \
-            .
+          make build
           zip -r -j out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_linux_${{ matrix.arch }}.zip dist/
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,99 @@
...
+          GOOS: "linux"
+          GOARCH: ${{ matrix.arch }}
+        run: |
+          mkdir dist out
+          CGO_ENABLED=0 go build \
+            -ldflags "-X github.com/hashicorp/vault-csi-provider/internal/version.BuildVersion=$(make version) \
+              -X github.com/hashicorp/vault-csi-provider/internal/version.BuildDate=$(date +%Y-%m-%d-%H:%M) \
+              -X github.com/hashicorp/vault-csi-provider/internal/version.GoVersion=$(go version | cut -d' ' -f3)" \
+            -o dist/ \
+            .
```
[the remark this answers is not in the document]
tomhjp: Good point - fixed.

Makefile
```diff
@@ -18,7 +18,7 @@ CI_TEST_ARGS?=
 XC_PUBLISH?=
 PUBLISH_LOCATION?=https://releases.hashicorp.com
```
tvoran, line 19: Might be able to remove some of these vars if they're not being used anymore:
- GOLANG_IMAGE
- XC_PUBLISH
- PUBLISH_LOCATION

.release/ci.hcl
```diff
@@ -0,0 +1,96 @@
...
+project "vault-csi-provider" {
+  team = "vault"
+  slack {
+    // #vault-releases channel
+    notification_channel = "CRF6FFKEW"
+  }
+  github {
+    organization = "hashicorp"
+    repository = "vault-csi-provider"
+    release_branches = ["master"]
```
tvoran, line 12:
```suggestion
    release_branches = ["main"]
```

tvoran (approved):

.release/ci.hcl
tomhjp: Thanks for the catch - surprisingly this was the only mention that needed fixing after the switch.

Makefile
tomhjp: 👍 good spot, GOOS and GOARCH were actually unused too (double checked they weren't being used for the `build` target when run on non-linux).

PR #129, opened 2022-01-11
Tidy usage and documentation of arguments

tomhjp:
Comments and further tidying following from #127

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ e2e-test:
 	bats test/bats/provider.bats
 
 # Check the current behaviour of -write-secrets flag and switch it.
-# If the flag is missing, switch to false because the default is true.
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -99,10 +99,10 @@ e2e-teardown:
 e2e-test:
 	bats test/bats/provider.bats
 
-# Check the current behaviour of --write-secrets flag and switch it.
+# Check the current behaviour of -write-secrets flag and switch it.
 # If the flag is missing, switch to false because the default is true.
```
tvoran, line 103: The default is false now?

```suggestion
# If the flag is missing, switch to true because the default is false.
```

tvoran (approved):

Makefile
tomhjp: Good spot, thanks

PR #133, opened 2022-01-21
Deprecate -write-secrets and -health_addr flags

tomhjp:
Preparing for v1.0.0 by removing deprecated flags. These have previously been marked deprecated in the changelog and release notes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,4 +91,3 @@ jobs:
           arch: ${{matrix.arch}}
           tags: |
             docker.io/hashicorp/${{env.repo}}:${{env.version}}
-            ecr.public.aws/hashicorp/${{env.repo}}:${{env.version}}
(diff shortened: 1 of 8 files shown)


tvoran (approved):

.github/workflows/build.yml
```diff
@@ -90,5 +90,4 @@ jobs:
           target: default
           arch: ${{matrix.arch}}
           tags: |
-            docker.io/hashicorp/${{env.repo}}:${{env.version}}
-            ecr.public.aws/hashicorp/${{env.repo}}:${{env.version}}
```
tvoran, line 94: I'm assuming this is being handled elsewhere in the release pipeline now?

tomhjp: We've actually never published these images to ECR, and never intended to. Including this line was just a copy paste mistake from the base example.

PR #143, opened 2022-02-11
feat: add bunch of ldflags for reproducibility

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>
cc: @tomhjp 🙋🏻‍♂️


tomhjp:
Hi @developer-guy, please can you include some documentation and rationale in the Makefile for those options?

developer-guy:
Hello @tomhjp, I found this: 
```shell
- trimpath increases build reproducibility. Per default full file paths are added to the go binary so two users will very likely have different binaries build even though the code is the same.
- buildmode=pie Enable position-independent code
- w No DWARF (reduces binary size)
- s No symbol table (even less binary size)
- znow and zrelro Hardening, adds runtime costs. See https://www.redhat.com/en/blog/hardening-elf-binaries-using-relocation-read-only-relro
```
> https://giters.com/ansible/receptor/issues/398

developer-guy:
And also a bunch of useful articles about reproducible builds for Go binaries:

* https://hackernoon.com/building-reproducible-verifiable-binaries-with-golang
* https://words.filippo.io/reproducing-go-binaries-byte-by-byte/

developer-guy:
We reduced the binary size up to 4MB 📉
Screen Shot 2022-02-11 at 21 42 01
Screen Shot 2022-02-11 at 21 41 14

developer-guy:
> Generally speaking; go and build flags is a rabbit hole since Go is unable to parse cli flags and env flags correctly. There is also a difference between hardening flags (PIE and RELRO) and reproducible builds flags. Setting trimpath and unsetting buildid is enough.

thanks to @Foxboron

developer-guy:
kindly ping @tomhjp

tomhjp:
Hi @developer-guy, please can you put the reasoning in the code alongside/just before the usage?

* `-buildid=` - I'm not convinced the need for this one is proven. It's set to a hash of the build inputs - that could be stable. Unfortunately I don't have the bandwidth to run experiments around this one
* `-s -w` - I'm fine with this
* `-trimpath` LGTM
* `-mod=readonly` LGTM
* `-modcacherw` - needs more justification. What paths? And why is it important that they're writeable?

I'm happy to land this either with just the ones that are definitely ok, or you're welcome to argue your case :)

PR #144, opened 2022-02-14
feat: allow customizing the "aud" claim in the JWT

a-riva:
Allow optionally specifying the audience in the SecretProviderClass to allow a safe usage of the ["audience" parameter in the Vault Kubernetes auth method](https://www.vaultproject.io/api-docs/auth/kubernetes#:~:text=audience%20(string%3A%20%22%22)%20%2D%20Optional%20Audience%20claim%20to%20verify%20in%20the%20JWT.).
If a custom audience is specified, it is required to use as reviewer JWT in Vault something different from [the client JWT](https://www.vaultproject.io/docs/auth/kubernetes#use-the-vault-client-s-jwt-as-the-reviewer-jwt).

This PR was tested against GKE, using a [static token as reviewer JWT](https://www.vaultproject.io/docs/auth/kubernetes#continue-using-long-lived-tokens).

Example usage:
```yaml
apiVersion: secrets-store.csi.x-k8s.io/v1
kind: SecretProviderClass
metadata:
  name: vault-credentials-class
spec:
  provider: vault
  parameters:
    audience: my-custom-audience # <-- this is added by this PR
    roleName: my-kubernetes-role
    objects: |
      - objectName: "testSecret"
        secretPath: "kv/data/secretPath"
        secretKey: "secretKey"
```

TODO:
- [x] Update the CHANGELOG
- [ ] Update the [documentation pages](https://www.vaultproject.io/docs/platform/k8s/csi)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -68,6 +68,7 @@ type Parameters struct {
 	VaultTLSConfig           api.TLSConfig
 	Secrets                  []Secret
 	PodInfo                  PodInfo
+	Audience                 string
 }
 
 type PodInfo struct {
@@ -128,6 +129,7 @@ func parseParameters(parametersStr string) (Parameters, error) {
 	parameters.PodInfo.UID = types.UID(params["csi.storage.k8s.io/pod.uid"])
 	parameters.PodInfo.Namespace = params["csi.storage.k8s.io/pod.namespace"]
 	parameters.PodInfo.ServiceAccountName = params["csi.storage.k8s.io/serviceAccount.name"]
+	parameters.Audience = params["audience"]
 	if skipTLS, ok := params["vaultSkipTLSVerify"]; ok {
 		value, err := strconv.ParseBool(skipTLS)
 		if err == nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ IMPROVEMENTS:
   * `-vault-tls-client-cert`
   * `-vault-tls-client-key`
   * `-vault-tls-skip-verify`
+* Add an optional SecretProviderClass parameter `audience` to customize the `aud` claim in the JWT [[GH-144](https://github.com/hashicorp/vault-csi-provider/pull/144)]
 
 ## 1.0.0 (January 25th, 2022)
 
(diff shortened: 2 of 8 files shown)


ocardia:
⬇️  Super minor: even if it's a test configuration, consider changing `my-aud` into `my-audience`.

internal/config/config.go
```diff
@@ -68,6 +68,7 @@ type Parameters struct {
 	VaultTLSConfig           api.TLSConfig
 	Secrets                  []Secret
 	PodInfo                  PodInfo
+	Audience                 string
```
tomhjp, line 71: Might it make sense to make this a string slice instead so it matches the way its used in the k8s API better?

tomhjp (commented):
Thanks for the contribution! I'd love to get this merged and it looks great for the most part. In addition to the string slice question, I would love to see an example of the audience parameter getting used in one of the `bats` acceptance tests. I can help out with that if you'd like.

internal/config/config.go
a-riva: I'm not sure it is a good idea: while for the k8s API, in general, it makes sense to accept multiple audiences, the Vault Kubernetes auth method only accepts [a single audience for each role](https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/main/path_role.go#L49), which should be already known at this point of the configuration (`audience` and `roleName` are set together, at the same level).

Moreover, including more than one `aud` in the JWT claims increases the risk of misuse of the token, whereas having only Vault as the specified audience provides better security (and, minor, allows to spare 1 line in the SecretProviderClass configuration).

However, if you still see value in changing this to a string slice, of course I can update it, just let me know 😊

a-riva:
Thank you for taking care of reviewing this PR!

I added an e2e test; let me know if it matches your expectations 😄

internal/config/config.go
tomhjp: Good point! Yeah this is a single-purpose token so that makes complete sense.

tomhjp (commented):
The e2e test looks great, thanks! Please could you add a changelog entry under improvements? And then I think this is ready to land 🚀

tomhjp (approved):
Awesome, thanks! If you'd like to, the configuration docs are generated from this file: https://github.com/hashicorp/vault/blob/main/website/content/docs/platform/k8s/csi/configurations.mdx

Thanks for the contribution!

tomhjp:
> When do you expect this to be released?

Should be released within the next couple of weeks or so!

PR #158, opened 2022-05-03
Add release metadata config key

jeanneryan:
Add release metadata config key to staging workflow event.

### DIFF
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -109,6 +109,7 @@ event "promote-staging" {
     organization = "hashicorp"
     repository = "crt-workflows-common"
     workflow = "promote-staging"
+    config = "release-metadata.hcl"
   }
 
   notification {


.release/ci.hcl
```diff
@@ -109,6 +109,7 @@ event "promote-staging" {
     organization = "hashicorp"
     repository = "crt-workflows-common"
     workflow = "promote-staging"
+    config = "release-metadata.hcl"
```
tomhjp, line 112: Any reason not to do this for production as well?

tomhjp (commented):
Is this something we need to do for other CRT repos as well?

tomhjp (approved):

jeanneryan:
> Is this something we need to do for other CRT repos as well?

Hi @tomhjp ! Yes, we have a list of repos that we are updating with this change.

.release/ci.hcl
jeanneryan: Great question! @sarahethompson will be able to answer this. I will ping her.

sarahethompson: 👋 - The release metadata is only needed for the initial publish (to staging). During the `promote-production` workflow we do a (new) `hc-releases promote` which promotes the files and metadata from staging to prod.

PR #161, opened 2022-06-07
Update testify and YAML deps

kalafut:
Update deps to address recent [YAML package CVE](https://github.com/advisories/GHSA-hp87-p4gw-j4gq).


swenson (approved):
LGTM

Issue #90, opened 2021-05-05
Vault token (and credentials) active renew of lease during lifetime of Pod to avoid TTL expiration?


thomas-riccardi:
The vault-agent-in-sidecar does active renewal of token (and/ credentials) lease for the lifetime of the Pod.
cf https://www.vaultproject.io/docs/platform/k8s/injector#renewals-and-updating-secrets and https://www.vaultproject.io/docs/agent/template#renewals-and-updating-secrets

That's especially useful as it allows using workload without any modification: it can read the secrets as env var at startup as usual and never be bothered to dynamically rotate them.

I could not find anything like that in this secret csi provider. 
(the closest was https://github.com/hashicorp/vault-csi-provider/pull/64 talking about token bound to the pod lifetime, but as I understand it it's about kubernetes token, not vault token)

Am I missing something? Is it something that is planned to be added in the future?

Thanks!

tomhjp:
Hi @thomas-riccardi, thanks for the feedback! We are indeed planning something in this area. The likely route is that we will do #82 first, and then start deploying Vault Agent as a sidecar to the vault-csi-provider daemonset and route all requests though the Agent. Then we can use the same exact mechanism as the sidecar injector.

jonpjenkins:
To add, as alluded to by @thomas-riccardi , this would be really useful with regards to vault provided dynamic credentials, which can be renewable but would need to be handled by an agent.  Thanks much for the work!

jonpjenkins:
@tomhjp checking in on this - now that #138 closed #82 , is this work able to go forward? This feature would be really useful for our use case.

tomhjp:
@jonpjenkins yep I've been thinking about this quite a bit lately. Adding lease renewal should now be relatively straightforward, and I think it probably won't even require any further changes in this repo - it should all be deployment changes which we'll help automate in [hashicorp/vault-helm](https://github.com/hashicorp/vault-helm).

One slight wrinkle as far as caching is concerned though: the cache entry key depends on the Vault token, but because we generate a fresh k8s token on every mount request, the provider will never get a cache hit from Agent. That means when auto rotation is used (via the flag on the driver), a fresh secret will be fetched on every re-mount. I'm wondering about a new Agent feature to help fix this, which could perhaps rely on JWT validation because even if the token isn't stable, the identity is, so that could be a verifiably correct way to allow cache hits from different tokens. But I haven't explored that in depth yet.

LosAngeles971:
[...]

PR #163, opened 2022-06-14
Use token from CSI TokenRequests

tomhjp:
The Secrets Store CSI driver supports [sending a token](https://github.com/kubernetes-sigs/secrets-store-csi-driver/commit/7ac887a55786306157ad308c751dc1c1fc064fdf) as part of the mount request as of [1.1.0](https://github.com/kubernetes-sigs/secrets-store-csi-driver/releases/tag/v1.1.0) ([relevant issue with rollout timeline](https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/585)), which is based on this [feature](https://kubernetes-csi.github.io/docs/token-requests.html) from this [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1855-csi-driver-service-account-token/README.md).

TL;DR is you use `--set tokenRequests[0].audience="vault"` when installing the CSI driver, and then it will populate a `"csi.storage.k8s.io/serviceAccount.tokens"` field in the mount parameters which contains a token generated for the audience "vault". That way, `vault-csi-provider` doesn't have to generate its own token for the requesting pod, and we can remove those Kubernetes permissions once we're ready to deprecate that way of doing things. However, I don't think we should deprecate immediately, as it still relies on Kubernetes 1.21+ and a pretty recent CSI driver version.

A really nice feature is that it will cache the token per-pod, so each pod will get the same JWT on each mount request until it expires. This will make it much easier to add Vault Agent as a sidecar in a way that caching will work (i.e. issue #90).

### DIFF
--- a/test/bats/_helpers.bash
+++ b/test/bats/_helpers.bash
@@ -14,13 +14,16 @@ wait_for_success() {
 }
 
 setup_postgres() {
-    # Setup postgres
+    # Setup postgres, pulling the image first to help avoid CI timeouts.
+    POSTGRES_IMAGE="$(awk '/image:/{print $NF}' $CONFIGS/postgres.yaml)"
+    docker pull "${POSTGRES_IMAGE}"
+    kind load docker-image "${POSTGRES_IMAGE}"
     POSTGRES_PASSWORD=$(openssl rand -base64 30)
     kubectl --namespace=test create secret generic postgres-root \
         --from-literal=POSTGRES_USER="root" \
         --from-literal=POSTGRES_PASSWORD="${POSTGRES_PASSWORD}"
     kubectl --namespace=test apply -f $CONFIGS/postgres.yaml
-    kubectl wait --namespace=test --for=condition=Ready --timeout=5m pod -l app=postgres
+    kubectl wait --namespace=test --for=condition=Ready --timeout=10m pod -l app=postgres
 
     # Configure vault to manage postgres
     kubectl --namespace=csi exec vault-0 -- vault secrets enable database


internal/config/config.go
```diff
@@ -148,6 +149,29 @@ func parseParameters(parametersStr string) (Parameters, error) {
...
+		var tokens map[string]struct {
+			Token               string `json:"token"`
+			ExpirationTimestamp string `json:"expirationTimestamp"`
+		}
+		if err := json.Unmarshal([]byte(tokensJSON), &tokens); err != nil {
+			return Parameters{}, fmt.Errorf("failed to unmarshal service account tokens: %w", err)
+		}
+
+		audience := "vault"
+		if parameters.Audience != "" {
```
swenson, line 167: For my own reference: the audience just needs to be a service account that a token will be generated for?

swenson (approved):
LGTM

internal/config/config.go
tomhjp: The audience can be any arbitrary string, not necessarily a service account name. If the user wants to take advantage of CSI TokenRequest tokens from the driver, they'll need to set an audience that matches under `tokenRequests` in the CSI driver definition: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/templates/csidriver.yaml#L13 (i.e. `--set tokenRequests[0].audience="vault"` via helm). The driver will create and send a token for each entry in that list.

If there's no token matching our expected audience, we'll currently fall back to the existing behaviour of generating the token in the provider, but I'd like to deprecate that fall back when this driver feature is more stable.

The idea is then the consumer of the JWT (Vault in our case) will validate that the JWT was intended to be consumed by Vault - for Kubernetes auth, that expected audience is configurable per role: https://www.vaultproject.io/api-docs/auth/kubernetes#create-role.

I'll write a changelog note announcing the plan to deprecate as well as the feature itself.

swenson: Cool, thanks for the information.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 CHANGES:
 
 * Duplicate object names now trigger an error instead of silently overwriting files. [[GH-148](https://github.com/hashicorp/vault-csi-provider/pull/148)]
+* Vault CSI Provider will use service account tokens passed from the Secrets Store CSI Driver instead of generating one if an appropriate token is provided. [[GH-163](https://github.com/hashicorp/vault-csi-provider/pull/163)]
+  * **NOTE:** Generating service account tokens within the Vault CSI Provider will be deprecated in **1.4.0**.
```
swenson, line 7:
```suggestion
  * **NOTE:** Generating service account tokens is deprecated within the Vault CSI Provider and will be removed in **1.4.0**.
```
?

internal/config/config.go
tomhjp: One knock-on I only just thought of.. Defaulting to an audience that is not the Kubernetes' cluster's default audience means these tokens won't be usable for our "use the client's JWT for auth" k8s auth option: https://www.vaultproject.io/docs/auth/kubernetes#use-the-vault-client-s-jwt-as-the-reviewer-jwt

I think that's probably a good thing that's more in the spirit of JWTs/OIDC, because it means tokens are generated to be consumed by a single audience. And even if we did go down the default audience route, it wouldn't simplify the driver configuration - users would just have to do a slightly confusing `--set tokenRequests[0].audience=""`.

So if you agree, probably the only action item here is to document that subtle detail when we next release the CSI provider.

tomhjp:
Sorry for the long delay on this PR. I've backed out the changelog notice that token generation will be deprecated, as I think that may still be useful longer term in some scenarios, but otherwise just merged in main.

PR #166, opened 2022-07-18
Ensure we check for the key to exist before attempting to read it

### LINKED ISSUE #165: Provider writes `null` if key does not exist at the secret path
### Steps to reproduce

Mount a K/V secret at `/foo/bar` containing the following:

```json
{
  "baz": 42
}
```

### Create the following `SecretProviderClass`:

```yaml
apiVersion: secrets-store.csi.x-k8s.io/v1alpha1
kind: SecretProviderClass
metadata:
  name: my-spc
spec:
  provider: vault
  parameters:
    roleName: vault
    objects: |
      - objectName: "something-name"
        secretPath: "foo/bar"
        secretKey: "non-existing-key"
```

#### Expected:

The pod to not boot up since `foo/bar/non-existing-key` does not exist in vault. 

#### What is happening:

The `provider.keyFromData` fails to fail and returns a `null` value. Which it renders on the filesystem path as if the secret contains a concrete null value.

The pod continues to start with a false-positive match of the secret provider class mount path.

### ⚠️ High Risk:

Imagine an application relies on `/super/secret/password` path and a human operator creates a typo on that path and writes `passsword` instead. The CSI Provider will render `null` for the password of the application. This could lead to failure in setting an actual password.

sarahhenkens:
This PR fixed #165, ensuring we check for the key existence before we pass it into json.Marshal. 

`data[secretKey]` returns `nil` and `json.Marshal(nil)` hides the fact that the secretKey doesn't exist. 

```
➜ make test 
gotestsum --format=short-verbose 
PASS internal/version.TestGetVersion (0.00s)
PASS internal/version
PASS internal/config.TestParseParametersFromYaml (0.00s)
PASS internal/config.TestParseParameters (0.00s)
PASS internal/config.TestParseConfig (0.00s)
PASS internal/config.TestParseConfig_Errors (0.00s)
PASS internal/config.TestValidateConfig (0.00s)
PASS internal/config
PASS TestListen (0.00s)
PASS .
PASS internal/provider.TestEnsureV1Prefix (0.00s)
PASS internal/provider.TestGenerateRequest (0.00s)
PASS internal/provider.TestKeyFromData (0.00s)
PASS internal/provider.TestKeyFromDataMissingKey (0.00s)
PASS internal/provider.TestHandleMountRequest (0.12s)
PASS internal/provider
PASS internal/client.TestNew (1.81s)
PASS internal/client.TestConfigPrecedence (0.00s)
PASS internal/client.TestNew_Error (0.00s)
PASS internal/client
EMPTY internal/server

DONE 15 tests in 6.840s
```

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -225,7 +225,12 @@ func (p *provider) getSecret(ctx context.Context, client *api.Client, secretConf
 		return content, nil
 	}
 
-	return keyFromData(secret.Data, secretConfig.SecretKey)
+	value, err := keyFromData(secret.Data, secretConfig.SecretKey)
+	if err != nil {
+		return nil, fmt.Errorf("{%s}: {%s}", secretConfig.SecretPath, err)
+	}
+
+	return value, nil
 }
 
 // MountSecretsStoreObjectContent mounts content of the vault object to target path
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -227,7 +227,7 @@ func (p *provider) getSecret(ctx context.Context, client *api.Client, secretConf
 
 	value, err := keyFromData(secret.Data, secretConfig.SecretKey)
 	if err != nil {
-		return nil, fmt.Errorf("{%s}: {%s}", secretConfig.SecretPath, err)
+		return nil, fmt.Errorf("{%s}: {%w}", secretConfig.SecretPath, err)
 	}
 
 	return value, nil
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -161,6 +161,11 @@ func keyFromData(rootData map[string]interface{}, secretKey string) ([]byte, err
 		data = rootData
 	}
 
+	// Fail early if a the key does not exist in the secret
+	if _, ok := data[secretKey]; !ok {
+		return nil, fmt.Errorf("key %q does not exist at the secret path", secretKey)
+	}
+
 	// Special-case the most common format of strings so the contents are
 	// returned plainly without quotes that json.Marshal would add.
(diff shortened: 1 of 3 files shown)


internal/provider/provider.go
```diff
@@ -161,6 +161,11 @@ func keyFromData(rootData map[string]interface{}, secretKey string) ([]byte, err
 		data = rootData
 	}
 
+	// Fail early if a the key does not exist in the secret
+	if _, ok := data[secretKey]; !ok {
+		return nil, fmt.Errorf("key %q does not exist at the secret path", secretKey)
```
tomhjp, line 166: For cases where the SecretProviderClass has a larger number of secrets, it would be nice to provide a bit more detail. It will require passing more info into this function, but I think it's worth it to help users debug quicker. Maybe we should pass the whole `config.Secret` struct into this function, and include at least the object name in the error message, possibly the path as well?

tomhjp (commented):
This looks great, thanks for including so much detail and tests upfront! I just have one suggestion to include more detail in the error message, and please could you also add a short entry to the changelog? Then I think this will be ready to merge.

internal/provider/provider.go
sarahhenkens: Good idea!

Instead of changing the function of the `keyFromData` (and all of its test cases). I ended up just prefixing the returned errors from this function and handle the additional detail from `GetSecret`.

This keeps the key from data decoupled from the `SecretConfig`

tomhjp: Nice! That's much better

tomhjp (approved):
Hope you don't mind I made 3 characters worth of nit edits just to save another round of review :)

Thanks for contributing this! It looks great 👌

PR #169, opened 2022-08-04
Provide release version at build time, move tests to GitHub Actions

tomhjp:
Two changes to better align the CI and release process with our other repos:
- Provide CRT with the version to use at build time
- Move CircleCI tests to GitHub actions and use a matrix test strategy


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,143 @@
...
+        with:
+          cluster_name: kind
+          node_image: kindest/node:v${{ matrix.kind-k8s-version }}
+          version: ${{ env.KIND_VERSION }}
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: vault-csi-provider-image
+
+      - name: Load vault-k8s dev image
```
tvoran, line 134: vault-k8s? Or is this the csi provider image?

Makefile
```diff
@@ -20,24 +19,27 @@ K8S_VERSION?=v1.22.2
 CSI_DRIVER_VERSION=1.0.0
 VAULT_HELM_VERSION=0.16.1
 CI_TEST_ARGS?=
```
tvoran, line 21: Is `CI_TEST_ARGS` used anymore?

tvoran (approved):

.github/workflows/tests.yaml
tomhjp: Oops, thanks, this was a copy-paste error

Makefile
tomhjp: Good spot 👍

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,143 @@
...
+          version: ${{ env.KIND_VERSION }}
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: vault-csi-provider-image
+
+      - name: Load vault-csi-provider dev image
+        run: |
+          docker image load --input ${{ env.TARBALL_FILE }}
+          kind load docker-image e2e/vault-csi-provider:latest
```
benashz, line 138: Not sure, but it looks like  the `e2e-setup` make target is already loading the docker image here: https://github.com/hashicorp/vault-csi-provider/blob/c4a6bdf92e947f348ff339b135e15f5d1d2b080b/Makefile#L66

benashz:
Seems to be unrelated, but I noticed that the `csi` namespace is not being deleted  when on  bats test tear down. This prevents re-running the tests.

```
kubectl create namespace csi
Error from server (AlreadyExists): namespaces "csi" already exists
make: *** [e2e-setup] Error 1
```

benashz (approved):
Have a few minor suggestions, otherwise LGTM!

.github/workflows/build.yml
```diff
@@ -15,8 +36,9 @@ jobs:
       - name: get product version
         id: get-product-version
         run: |
-          make version
-          echo "::set-output name=product-version::$(make version)"
+          VERSION="${{ github.event.inputs.version || '0.0.0-dev' }}"
```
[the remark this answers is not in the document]
tomhjp: We never publish them, the main intention is to surface breakages of that pipeline quickly. But seems like a nice addition to include the SHA for the dev versions anyway.

PR #176, opened 2022-08-25
Update alpine and go.mod dependencies

tomhjp:
I also upgraded the go.mod file's Go version to get the nice concrete list of indirect dependencies.


tvoran:
Looks like the new k8s client needs go 1.19? https://github.com/hashicorp/vault-csi-provider/runs/8026118782?check_suite_focus=true

So I guess we should move to 1.19 or bump the k8s libraries back a rev.

Interesting that it's complaining about RawRequestWithContext too: https://github.com/hashicorp/vault-csi-provider/runs/8026118782?check_suite_focus=true

Dockerfile
```diff
@@ -1,6 +1,8 @@
 # This Dockerfile contains multiple targets.
 # Use 'docker build --target=<name> .' to build one.
 
+ARG ALPINE_VERSION=3.16.2
+
 # devbuild compiles the binary
 # -----------------------------------
 FROM docker.mirror.hashicorp.services/golang:1.17.6 AS devbuild
```
tvoran, line 8: Seems like this golang version should be changed too.

tvoran: Actually, I think this is why the tests are failing; settings this to 1.18.5 passes `make e2e-container` for me.

Dockerfile
```diff
@@ -1,9 +1,12 @@
 # This Dockerfile contains multiple targets.
 # Use 'docker build --target=<name> .' to build one.
 
+ARG ALPINE_VERSION=3.16.2
+ARG GO_VERSION
+
 # devbuild compiles the binary
 # -----------------------------------
-FROM docker.mirror.hashicorp.services/golang:1.17.6 AS devbuild
+FROM docker.mirror.hashicorp.services/golang:${GO_VERSION} AS devbuild
```
tvoran, line 9: Will this work with the CRT docker build? I think docker build may complain if the image tag isn't set.

tomhjp: Good point, I thought I had successfully run `docker build --target=default .` to test it before pushing, but it did break CRT builds. I've defaulted it to `latest` now, just so that the Dockerfile is valid and can be parsed, but that latest tag isn't actually used in any Makefile targets or CRT builds.

tvoran (approved):

PR #182, opened 2022-10-25
Try to fix security issues

jdanielgg:
# Issue to solve security problems

We have been used your repository, however, the vulnerabilities test in [quay.io ](https://quay.io/repository/urbanos/vault-csi-provider/manifest/sha256:b0ea7602b01ccaab13e1575bb51a3534f4b9f04971c94c2db664bcab1322e932?tab=vulnerabilities) shows some critical security issues. 

We added the "RUN apk update && apk upgrade" command in order to fix it and the results are succesfull.

Could you review it?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,8 @@ ARG PRODUCT_REVISION
 ARG PRODUCT_NAME=vault-csi-provider
 ARG TARGETOS TARGETARCH
 
+RUN apk update && apk upgrade
+
 LABEL version=$PRODUCT_VERSION
 LABEL revision=$PRODUCT_REVISION
 


tomhjp:
Hi @jdanielgg, thanks for your interest in fixing this! Reading up on the critical CVE, it seems to be in `inflate.c`, and `vault-csi-provider` doesn't use `cgo`, so I don't think it should be practically exploitable. Even if my analysis is correct though, I do agree that it's undesirable to have false positives getting flagged. Having said that, I'm hesitant to accept this patch as-is because I would prefer not to make the build less reproducible.

I'd be open to pinning to an upgraded package version until there's a new alpine image available. Would `apk add alpine-baselayout=3.2.0-r23` do the trick?

It's worth noting if you have any new security issues you'd like to raise in the future, we do encourage emails to security@hashicorp.com for private disclosure: https://github.com/hashicorp/.github/blob/master/SECURITY.md. But I'm happy to discuss this one on GitHub with my current understanding.

tomhjp:
1.2.1 just got released, which is built on alpine 3.16.3, and should not have this vulnerability. I'm going to close this for now, but I still think we could make some improvements so I do plan to revisit this topic. Thank you for raising this!

Issue #106, opened 2021-08-24
Support for secret encoding?


schantaraud:
It seems that a common way of storing non-Unicode secrets in Vault is to base64-encode them. Is there currently a way to mount these through the CSI provider without adding a manual decoding process?
If not, the Azure driver supports an `objectEncoding` property (see https://github.com/Azure/secrets-store-csi-driver-provider-azure/pull/236). Would you accept a PR that adds a similar feature?

bimtauer:
Same here, this would be great to have. We run some third party services that require certificates in pem string format but we have them base64 encoded everywhere else in Vault.

LianwMS:
@therealsamlin and @bimtauer, then how did you resolve such issue? We have the same issue here.

tomhjp:
Fixed by #194. Thanks for the contribution @BroCanDo!

PR #194, opened 2023-02-12
Support Encoding of Secrets

BroCanDo:
Added support for encoded secrets (#106). This makes sense e.g. in the context of non-UTF-8 encoded certificates, like pfx.
This would allow storing e.g. 
* asymmetric private/public key pairs for Pulsar
* certificates for Elasticsearch

as base64 encoded strings in Vault.

Since this is my first go with go (pun intended) I would appreciate any feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+CHANGES:
+
+* Support utf-8 (default), hex, and base64 encoded secrets
+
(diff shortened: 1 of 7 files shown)


tomhjp (approved):
Sorry for the delay reviewing - this is super thorough, thanks for including all the tests! LGTM with one minor lint failure that I just pushed a trivial fix for, hope you don't mind. I'll merge once the CI goes green.

PR #197, opened 2023-03-20
Fix changelog for #163 to be unreleased

tomhjp:
I messed up the merging in of main on #163, it hasn't been released yet as it only just reached main.


swenson (approved):
LGTM

PR #198, opened 2023-03-24
Update secret version hashing algorithm

tomhjp:
Updates the version generation algorithm to include an HMAC key. On each mount request, the provider will try to read the HMAC key from a Kubernetes secret and race to create it if not found. This ensures each provider produces consistent versions, and also makes recovering from unexpected errors easy (an admin just deletes the secret) without introducing the complexity and overhead of leader elections.

The PR also makes generating versions best-effort. If we can't use an HMAC key, we log a warning and revert to our pre-1.2.0 behaviour of not reporting a version at all, as consistency and reliability seem much more important than accurate version reporting. If versions thrash about unnecessarily it will cause lots of thrashing for any systems that observe the version.

Sorry for the size. I also made a feature to specify custom metadata for the created secret in https://github.com/hashicorp/vault-csi-provider/commit/4b422ed4cc8ba6f750b1401029e6b0765f2b33ea, but felt it wasn't quite straightforward enough so took that feature out for the initial PR to keep the size lower. The idea was to replicate the pod's metadata, but there are certain pod-specific labels you'd want to filter out like `controller-revision-hash` and `pod-template-generation`. I'll try to refine that idea and follow up in another PR.


internal/provider/provider.go
```diff
@@ -311,12 +319,21 @@ func (p *provider) HandleMountRequest(ctx context.Context, cfg config.Config, fl
...
+			Version: "",
+		}, nil
+	}
+
 	// We include the secret config in the hash input to avoid leaking information
 	// about different secrets that could have the same content.
-	_, err := hash.Write([]byte(fmt.Sprintf("%v:%s", secret, content)))
-	if err != nil {
+	hash := hmac.New(sha256.New, hmacKey)
+	if _, err := hash.Write([]byte(fmt.Sprintf("%v:%s", secret, content))); err != nil {
```
swenson, line 336: While we're here changing the hash, I think it might be prudent to use something with other than `%v` to serialize the secret, like JSON or gob or `encoding/binary`, so that it won't contain, say, memory addresses and has more guarantees about consistent encoding.

swenson (approved):
LGTM; one suggestion

main.go
```diff
@@ -95,16 +105,30 @@ func realMain(logger hclog.Logger) error {
...
 	}
 
-	srv := providerserver.NewServer(serverLogger, flags, clientset)
+	namespace, err := os.ReadFile(namespaceFile)
+	if err != nil {
+		return fmt.Errorf("failed to read namespace from file: %w", err)
+	}
+	hmacSecretSpec := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      *&flags.HMACSecretName,
```
tvoran, line 123: Does this need to be dereferenced and then referenced for some reason?

```suggestion
			Name:      flags.HMACSecretName,
```

tvoran (approved):
The reactor usage in the tests is really nice!

main.go
tomhjp: Oops, thanks for spotting that.

PR #200, opened 2023-04-05
Prepare 1.3.0 release

tomhjp:
Prepare for 1.3.0 release:

* Changelog
* Update deps:
  * Update & pin GHA, point jira sync at common workflows
  * Go version
  * Go mod
  * Docker base image
  * CSI driver version in tests
  * CI tool and k8s testing versions
* Run `make promote-staging-manifest`

Also fixed a small bug in the staging manifest's role permissions, and added one extra debug log line that helped me while testing out the latest changes.


.github/workflows/tests.yaml
```diff
@@ -107,11 +70,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kind-k8s-version: [1.21.14, 1.22.15, 1.23.13, 1.24.7, 1.25.3]
+        kind-k8s-version: [1.22.17, 1.23.17, 1.24.12, 1.25.8, 1.26.3]
```
kschoche, line 73: Are we deprecating support for 1.21? Do we need to document that, or that we won't test against it anymore?

go.mod
```diff
@@ -3,88 +3,70 @@ module github.com/hashicorp/vault-csi-provider
 go 1.19
```
tvoran, line 3: May want to bump this too?

```suggestion
go 1.20
```

tvoran (approved):

.github/workflows/tests.yaml
tomhjp: Good point, it's no longer supported by AKS, EKS, GKE, or the Kubernetes team, so I think we should drop it. I'll follow up with a docs PR for that.

tomhjp: Opened https://github.com/hashicorp/vault/pull/20010

PR #201, opened 2023-04-05
Update integration tests to use latest Vault

tomhjp:
The integration tests were still using a very old version of Vault Enterprise that allowed users to run it for a limited time without a license. Added a license to the repo so we can test a newer version of Vault, and also added some conditionals so that OSS users can run most of the tests themselves without requiring a license.

### DIFF
--- a/test/bats/provider.bats
+++ b/test/bats/provider.bats
@@ -13,13 +13,15 @@ CONFIGS=test/bats/configs
 setup(){
     { # Braces used to redirect all setup logs.
     # 1. Configure Vault.
-    kubectl --namespace=csi exec vault-0 -- vault namespace create acceptance
 
     # 1. a) Vault policies
     cat $CONFIGS/vault-policy-db.hcl | kubectl --namespace=csi exec -i vault-0 -- vault policy write db-policy -
     cat $CONFIGS/vault-policy-kv.hcl | kubectl --namespace=csi exec -i vault-0 -- vault policy write kv-policy -
     cat $CONFIGS/vault-policy-pki.hcl | kubectl --namespace=csi exec -i vault-0 -- vault policy write pki-policy -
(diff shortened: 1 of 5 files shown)


test/bats/provider.bats
```diff
@@ -30,16 +32,12 @@ setup(){
...
-        token_reviewer_jwt="$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
-        kubernetes_host="https://$KUBERNETES_PORT_443_TCP_ADDR:443" \
-        kubernetes_ca_cert=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
-        issuer="https://kubernetes.default.svc.cluster.local"'
+        kubernetes_host="https://$KUBERNETES_PORT_443_TCP_ADDR:443"'
+    if [ -n "${VAULT_LICENSE}" ]; then
+        kubectl --namespace=csi exec vault-0 -- vault auth enable -namespace=acceptance kubernetes
+        kubectl --namespace=csi exec vault-0 -- sh -c 'vault write -namespace=acceptance auth/kubernetes/config \
+            kubernetes_host="https://$KUBERNETES_PORT_443_TCP_ADDR:443"'
+    fi
```
tvoran, line 40: Does the `-namespace` option accept an empty value? Since the OSS and ENT commands are the same here, I wonder if we could set the namespace in a variable above somewhere if $VAULT_LICENSE is present, and otherwise leave it blank and run the same commands for both ENT and OSS.

tvoran (approved):

test/bats/provider.bats
tomhjp: It should do, however, for the ent tests we use both mounts. Most of the SPCs refer to non-namespaced secrets and then we have one SPC that uses a Vault namespace, so this is an additional mount for that test.

I would love to do a large refactor on these tests to be smaller and easier to work with, but I would also like to make them Go-based like some of our other newer integration test suites, so I was planning to save that big refactor for doing the switch at some point.

PR #207, opened 2023-04-25
Revert "Add copyright headers CI check"

tomhjp:
I accidentally pushed 3fb01aacd580878208ef3c8764f2ffce932b6bc5 straight to main with admin privilege. I've updated the repo settings to ensure rules can't be bypassed by admins and confirmed I wasn't able to push this revert directly to main:

```text
$ git push
Enumerating objects: 15, done.
Counting objects: 100% (15/15), done.
Delta compression using up to 10 threads
Compressing objects: 100% (7/7), done.
Writing objects: 100% (8/8), 2.13 KiB | 2.13 MiB/s, done.
Total 8 (delta 5), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (5/5), completed with 5 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: error: At least 1 approving review is required by reviewers with write access.
To github.com:hashicorp/vault-csi-provider.git
 ! [remote rejected] main -> main (protected branch hook declined)
error: failed to push some refs to 'github.com:hashicorp/vault-csi-provider.git'
```

Once reverted, I'll re-open a lightly modified version for proper review.

PR #208, opened 2023-04-25
Add copyright header CI checks

tomhjp:
Adds CI from the common repo for checking for copyright headers to reduce the number of PRs like #206. Also switches a couple of CI checks to use the common repo's version for consistency.

Don't review until https://github.com/hashicorp/vault-csi-provider/pull/207 is merged to main and pulled in, or the diff will be confusing.

I'll also wait until https://github.com/hashicorp/vault-workflows-common/pull/24 merges so that I can update the job to reference `main` instead of `add-copyright-headers-check`.


kschoche (approved):
Looks good!

PR #210, opened 2023-04-26
Support JWT auth

tomhjp:
Someone pointed out that JWT auth already works because the login endpoints are identical for [kubernetes](https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#login) and [JWT](https://developer.hashicorp.com/vault/api-docs/auth/jwt#jwt-login). This just improves the naming and adds an integration test for the purpose of regression testing so that we can officially document JWT auth as supported.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -70,14 +70,14 @@ func (fc FlagsConfig) TLSConfig() api.TLSConfig {
 //
 // So we just deserialise by hand to avoid complexity and two passes.
 type Parameters struct {
-	VaultAddress             string
-	VaultRoleName            string
-	VaultKubernetesMountPath string
-	VaultNamespace           string
-	VaultTLSConfig           api.TLSConfig
-	Secrets                  []Secret
-	PodInfo                  PodInfo
-	Audience                 string
+	VaultAddress       string
+	VaultRoleName      string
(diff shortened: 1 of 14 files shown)


kschoche (approved):
👏🏽 Looks great, thanks!

internal/config/config.go
```diff
@@ -136,7 +136,16 @@ func parseParameters(parametersStr string) (Parameters, error) {
 	parameters.VaultTLSConfig.TLSServerName = params["vaultTLSServerName"]
 	parameters.VaultTLSConfig.ClientCert = params["vaultTLSClientCertPath"]
 	parameters.VaultTLSConfig.ClientKey = params["vaultTLSClientKeyPath"]
-	parameters.VaultKubernetesMountPath = params["vaultKubernetesMountPath"]
+	k8sMountPath, k8sSet := params["vaultKubernetesMountPath"]
+	authMountPath, authSet := params["vaultAuthMountPath"]
+	switch {
```
kschoche, line 141: I've never seen this pattern of using switch statements this way in golang before, really cool!

kschoche (approved):
🤝

tomhjp:
Thanks!

PR #211, opened 2023-04-28
Prepare 1.4.0 release

tomhjp:
## Changes

* SecretProviderClass objects now also accept `spec.parameters.vaultAuthMountPath` as an alternative to `spec.parameters.vaultKubernetesMountPath`. [[GH-210](https://github.com/hashicorp/vault-csi-provider/pull/210)]

## Features

* The Provider will cache a Vault token per requesting pod in memory and re-use it until it expires. [[GH-202](https://github.com/hashicorp/vault-csi-provider/pull/202)]
* JWT auth is supported by setting role name and auth mount path in the same way as for Kubernetes auth. [[GH-210](https://github.com/hashicorp/vault-csi-provider/pull/210)]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+# 1.4.0 (April 28th, 2023)
+
 CHANGES:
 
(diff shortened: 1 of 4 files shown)


kschoche (approved):
⚡

tomhjp:
Thanks!

PR #234, opened 2023-12-05
feat(logging): Add json logging configuration

### LINKED ISSUE #177: Enable json logging format
Specifically for GKE environments that leverage Google Cloud Operations Logging, logs generated by the provider are all classified as ERROR rather than the proper type as outlined in stdout/stderr. It would be helpful to be able to set the log output in json format such that Operations Logging can properly classify the log type. 

image

jbgcarnes:
Closes #177 

Added json logging flag to allow logging in JSON format. Flags set to match Vaults flag `-log-format` and to support the same value `json`.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -47,6 +47,8 @@ type FlagsConfig struct {
 	TLSClientCert  string
 	TLSClientKey   string
 	TLSSkipVerify  bool
+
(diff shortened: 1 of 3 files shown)


shaafekhan:
Hi all. Any updates on this PR? Looking forward to being able to set log format to JSON.

PR #240, opened 2024-03-11
Test against k8s 1.29

benashz:
- Add k8s 1.29.2 to the k8s version test matrix. 
- Bump all previous minor versions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -55,7 +55,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


tomhjp (approved):
I'd be happy for us to also delete 1.24 at the same time - that's onto extended support with the major CSP hosted offerings.

benashz:
> I'd be happy for us to also delete 1.24 at the same time - that's onto extended support with the major CSP hosted offerings.

Dropped with 79d2919cf48eb79dbf675f94bf6b73b4fd018815

PR #242, opened 2024-03-12
Setup dependabot for the project.

benashz:


.github/dependabot.yaml
```diff
@@ -0,0 +1,24 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    # Disable version updates and only check security updates for github
+    # actions, since we can't bump the versions until they're on our allow-list
+    open-pull-requests-limit: 0
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
tomhjp, line 24: Any reason not to match the cadence for the other 2?

tomhjp (approved):
Thanks! 👍

.github/dependabot.yaml
benashz: Good question. I borrowed this from VSO. I think your suggestion makes sense to match them. I would imagine that package updates are more frequent than docker image updates.

benashz: Updated in bcde91bfd28ea5b0cad17012363b24cec6d13ed6

.github/dependabot.yaml
```diff
@@ -0,0 +1,24 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    # Disable version updates and only check security updates for GitHub
+    # actions, since we can't bump the versions until they're on our allow-list
+    open-pull-requests-limit: 0
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
tvoran, line 24: FWIW I don't think this will do much until the Dockerfile uses explicit tags in the FROM lines for dev and default. The indirection with $ALPINE_VERSION fools dependabot.

.github/dependabot.yaml
tvoran: (Just to add context here) IIRC the motivation for daily docker updates on VSO and vault-k8s was to catch ubi image updates, since RedHat has their own security scanning criteria that our release process wouldn't catch before pushing to RedHat's registry. CSI doesn't have a ubi image so 👍

.github/dependabot.yaml
benashz: Thanks. Put up #248  to address that.

PR #248, opened 2024-03-12
Inline the alpine image version for dependabot

benashz:


tomhjp (approved):
LGTM, I think it was only an argument to avoid repetition within the file

tvoran (approved):

PR #249, opened 2024-03-13
CI: test against n-2 versions of Vault

benashz:
This PR add support for testing the CSI provider against n-2 versions of Vault. It is adapted from the work done in the vault-secrets-operator project.

Relates to https://github.com/hashicorp/vault-k8s/pull/604


.github/workflows/tests.yaml
```diff
@@ -48,15 +48,74 @@ jobs:
           name: vault-csi-provider-image
           path: ${{ env.TARBALL_FILE }}
 
+  vault-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "setting vault versions"
+    outputs:
```
tomhjp, line 55: I don't mind this way, but the top-level `env` would probably be a better fit for setting these workflow constants instead of running a dedicated step to get outputs, ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env. You would then refer to the values similarly below as `${{ env.VAULT_N }}` if you need to use GitHub workflows interpolation to make it part of the step definition, or obviously you can also use the normal shell env syntax within shell scripts.

The main benefit you'd get is a bit of simplification because you no longer need to manage `needs` entries for these constants.

.github/workflows/tests.yaml
```diff
@@ -48,15 +48,74 @@ jobs:
...
+          - "${{ needs.k8s-versions.outputs.K8S_1_25 }}"
+          - "${{ needs.k8s-versions.outputs.K8S_1_26 }}"
+          - "${{ needs.k8s-versions.outputs.K8S_1_27 }}"
+          - "${{ needs.k8s-versions.outputs.K8S_1_28 }}"
+          - "${{ needs.k8s-versions.outputs.K8S_1_29 }}"
+        vault-version:
+          - "${{ needs.vault-versions.outputs.VAULT_N_2 }}"
+          - "${{ needs.vault-versions.outputs.VAULT_N_1 }}"
+          - "${{ needs.vault-versions.outputs.VAULT_N }}"
+        # Note: We want to test the operator with the different k8s versions,
```
tomhjp, line 92:
```suggestion
        # Note: We want to test the provider with the different k8s versions,
```

.github/workflows/tests.yaml
```diff
@@ -48,15 +48,74 @@ jobs:
...
+        # and with the different vault versions, but we don't care about testing
+        # all the k8s versions against all the vault versions.
+        # Combos to exclude:
+        #   kind-k8s-version: [K8S_1_25, K8S_1_26, K8S_1_27, K8S_1_28]
+        #   vault-version: [VAULT_N_2, VAULT_N_1]
+        exclude:
+          # vault version N-2
+          - kind-k8s-version: ${{ needs.k8s-versions.outputs.K8S_1_28 }}
```
tomhjp, line 100: Similarly to above, I am not too opinionated on this, but one improvement we could make would be to extract the testing steps below into a reusable action and call it with 2 different matrices. In one, we would specify the latest Vault version and all k8s versions, and in the other we would specify the latest k8s version and all Vault versions. Then we could get rid of this repetitive/mechanical list of exclusions.

tomhjp (approved):
LGTM 👍 I made some optional suggestions but this PR is a good improvement with or without them.

.github/workflows/tests.yaml
tvoran: IIRC we ran into trouble referring to top-level `env`'s in the matrix definition while setting this up for the operator. Maybe something's changed since then?

tomhjp: Oops sorry you're right, it seems env values can only be interpolated in steps, not jobs. The docs aren't terribly explicit about the jobs side of that, but here's the reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env

benashz:
@tomhjp I pulled in the relevant changes from your branch tomhjp/testing. Made some modifications as well. If you could make another pass that would be 🍠

tomhjp (approved):
Nice! 👍

.github/workflows/tests.yaml
```diff
@@ -48,41 +56,51 @@ jobs:
           name: vault-csi-provider-image
           path: ${{ env.TARBALL_FILE }}
 
-  integration-test:
+  latest-vault:
+    name: latest-vault vault:${{ matrix.vault-version }} kind:${{ matrix.k8s-version }}
```
tomhjp, line 60: nittiest of nitpicks, maybe omit the latest version so the job name fits in the github UI?

```suggestion
    name: latest-vault kind:${{ matrix.k8s-version }}
```

benashz: I dropped the job prefix here:  08c4e68 - I think that looks better now in the UI.

PR #295, opened 2024-07-29
Add ability to tune log levels

Zlaticanin:
This PR involves deprecating the current way we set a log level and replacing it with a new flag `-log-level`. The log level options include: info, debug, trace, warn, error, and off.

Before: `-debug=true`
Now: `-log-level=debug`

PR to update tests in https://github.com/hashicorp/vault-helm/pull/1048


main_test.go
```diff
@@ -32,3 +33,29 @@ func TestListen(t *testing.T) {
 	_, err = os.Create(endpoint)
 	require.NoError(t, err)
 }
+
+func TestSetupLogger(t *testing.T) {
+	tests := []struct {
+		flags    config.FlagsConfig
+		expected hclog.Level
+	}{
+		{config.FlagsConfig{Debug: true}, hclog.Debug}, // deprecated flag test
```
tvoran, line 42: Might be good to test the case where both `Debug` and `LogLevel` are set.

tvoran (approved):
Tested great for me locally 👍 

We should update the logging [args in `manifest_staging/deployment/vault-csi-provider.yaml`](https://github.com/hashicorp/vault-csi-provider/blob/c855c9411768a6dfbf1b3bd807d5ec1c1df8b584/manifest_staging/deployment/vault-csi-provider.yaml#L95) as well.

And feel free to add an entry to the changelog in the unreleased section for this.

main_test.go
Zlaticanin: Good idea. And this brings up a good point too - Do we want the `LogLevel` to take precedence if they are both set?
Currently, if the `Debug` flag is set to true and `LogLevel` is set to `warn` for example, the `Debug` flag will take precedence. This is how the `setupLogger` func is right now. Do we want to keep it like this, or change our `else` block. Do we have a preference?

tvoran: I'd probably expect `LogLevel` to take precedence now that you mention it, but I don't have a strong preference.

Zlaticanin: I agree, I think that LogLevel should take precedence. Updated in https://github.com/hashicorp/vault-csi-provider/pull/295/commits/eb12344a3d36e8d6b16efb39990f5d896157c7cf

PR #324, opened 2024-11-21
CI and build updates

tvoran:
Update golang to 1.23.4, `golang/x` packages, CI tools, and the image in the deployment yaml.

The go update will help with https://github.com/hashicorp/vault-csi-provider/pull/331, since grpc@v1.68.0 needs >= go 1.22.7.


.github/workflows/tests.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
         run: make bootstrap
 
       - name: Lint
-        run: make lint GOLANGCI_LINT_FORMAT=github-actions
+        run: make lint GOLANGCI_LINT_FORMAT=colored-line-number
```
benashz, line 38: Is this change required?

benashz (approved):
One question, otherwise +1

.github/workflows/tests.yaml
tvoran: I think it still works with a warning, but the "github-actions" format was deprecated in [v1.59.0](https://github.com/golangci/golangci-lint/releases/tag/v1.59.0)

PR #382, opened 2025-08-14
Building and testing a UBI-based image

tvoran:
This PR adds a [UBI-based](https://catalog.redhat.com/en/software/base-images) vault-csi-provider image to the Dockerfile and CI build workflow. This will produce a UBI image of vault-csi-provider for arm64 and amd64 architectures, which will be pushed to docker and redhat container registries at release time.

The bats integration tests and Makefile setup were also modified to run the tests against the UBI image on both kind and OpenShift. Instructions for running tests on OpenShift are in the README, and the CI tests workflow smoke tests the UBI image with the latest vault and k8s combo.

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.


test/bats/configs/nginx/templates/nginx.yaml
```diff
@@ -14,7 +14,7 @@ spec:
   serviceAccountName: nginx-{{ .Values.sa }}
   terminationGracePeriodSeconds: 0
   containers:
-  - image: docker.mirror.hashicorp.services/nginx
+  - image: docker.mirror.hashicorp.services/nginxinc/nginx-unprivileged
```
tvoran, line 17: The unprivileged nginx container is a lot easier to work with on OpenShift, and we don't need a privileged version for the tests, so I switched everything to the unprivileged version.

manifest_staging/deployment/vault-csi-provider.yaml
```diff
@@ -126,6 +126,6 @@ spec:
       volumes:
         - name: providervol
           hostPath:
-            path: "/etc/kubernetes/secrets-store-csi-providers"
+            path: "/var/run/secrets-store-csi-providers"
```
tvoran, line 129: "/etc/kubernetes/secrets-store-csi-providers" is the old provider volume folder, and "/var/run/secrets-store-csi-providers" has been the [default for a while](https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/870). Sometimes the `/etc` one is set as an additional provider volume for backwards compatibility, but not always (like with the CSI Driver operator in OpenShift).

test/bats/configs/vault-all-secretproviderclass.yaml
```diff
@@ -9,6 +9,7 @@ spec:
   provider: vault
   parameters:
     roleName: "all-role"
+    audience: "vault"
```
tvoran, line 12: Matches the audience on the vault role config.

test/bats/configs/vault-kv-secretproviderclass-jwt-auth.yaml
```diff
@@ -11,8 +11,9 @@ spec:
   parameters:
     roleName: "jwt-kv-role"
     vaultAuthMountPath: "jwt"
+    audience: "vault"
     objects: |
       - objectName: "secret-1"
         secretPath: "secret/data/kv1"
         secretKey: "bar1"
-        filePermission: 0600
+        filePermission: 0644
```
tvoran, line 19: Changed the file permissions in some of the tests since the OpenShift UBI image is running as a non-root user, but when the secrets store CSI driver writes files they're always owned by root: https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/858

test/bats/provider.bats
```diff
@@ -75,7 +75,7 @@ setup(){
     # 1. b) ii) Setup JWT auth
     kubectl --namespace=csi exec vault-0 -- vault auth enable jwt
     kubectl --namespace=csi exec vault-0 -- vault write auth/jwt/config \
-        oidc_discovery_url=https://kubernetes.default.svc.cluster.local \
+        oidc_discovery_url="$(kubectl get --raw /.well-known/openid-configuration | jq -r '.issuer')" \
```
tvoran, line 78: Made this dynamic since it'll differ between kind and OpenShift.

Dockerfile
```diff
@@ -51,6 +51,48 @@ COPY LICENSE /usr/share/doc/$PRODUCT_NAME/LICENSE.txt
...
+FROM registry.access.redhat.com/ubi10/ubi-micro:10.0-1754556444 AS release-ubi
+
+ENV BIN_NAME=vault-csi-provider
+ARG PRODUCT_VERSION
+ARG PRODUCT_REVISION
+ARG PRODUCT_NAME=$BIN_NAME
+# TARGETARCH and TARGETOS are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+LABEL name="Vault CSI Provider" \
```
digivava, line 70: This isn't a blocker, as I don't think it's the end of the world if this label says "Vault CSI Provider", but I think we decided that in order to more easily differentiate from our own CSI driver, we'd start calling this offering "Secrets Store CSI Provider for Vault" in more official human-readable contexts (i.e. we don't need to change the binary name/repo name). 

My PR for the VSO CSI driver docs replaces "Vault CSI Provider" with that name across all docs pages that mention it. It just makes it easier to understand that this is our contribution to someone else's CSI driver, while we also have our own.

May want to double-check this with Steve though, if you have any hesitation about us changing the official name!

digivava: The summary/description for the image just below may also cover this confusion 🤷‍♀️

digivava (dismissed):
I didn't run it myself in OpenShift to test (let me know if that's what you were hoping for) but I looked over it and it seems like you've got all the bases covered! Thanks for getting all this figured out!

.go-version
```diff
@@ -1 +1 @@
-1.24.5
+1.24.6
```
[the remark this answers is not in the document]
tvoran: Yeah I don't know of any need to use 1.25.0 yet. And this change was specifically to pass the security scan so the artifacts would be generated properly.

Dockerfile
```diff
@@ -51,6 +51,48 @@ COPY LICENSE /usr/share/doc/$PRODUCT_NAME/LICENSE.txt
...
+      maintainer="Vault Team <vault@hashicorp.com>" \
+      vendor="HashiCorp" \
+      version=$PRODUCT_VERSION \
+      release=$PRODUCT_VERSION \
+      revision=$PRODUCT_REVISION \
+      org.opencontainers.image.licenses="BUSL-1.1" \
+      summary="HashiCorp Vault Provider for Secret Store CSI Driver for Kubernetes" \
```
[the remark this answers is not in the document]
tvoran: Thanks, I think this came from `About` on the repo.

Dockerfile
tvoran: Ok, these metadata labels are just copied from the existing alpine image above. If we're adjusting the name I think we'll also want to update the descriptions in the docker hub and Red Hat registries.

Side note, "Secrets Store CSI Provider for Vault" sounds a little backwards to me? Should it be something more like Vault Secrets Store CSI Provider? i.e. Vault is the thing being provided, not provided for in my mind.

